PR #813, opened 2016-02-17
OTWO-4134: Replace contributions view with materialized view - DON'T MERGE!

PDegenPortnoy:
This PR will fail because code coverage is 99.52% instead of 99.55%.  I'm working on it by adding some test somewhere, but it seems silly to hold up the PR because of that.


PDegenPortnoy:
We want to hang onto this work, but we don't need the PR. We definitely want to keep the branch

PR #1155, opened 2017-05-22
OTWO-4770: Improve performance on commits controller index page

santhanakarthikeyan:


PDegenPortnoy:
OK to merge even with some broken tests (due to the challenge in FDW and what is captured in structure.sql)

PR #1162, opened 2017-05-26
Fixed Admin#Dashboard performance issue

santhanakarthikeyan:
I've commented out 'Oldest Repository' code as it is a performance intensive query. 

explain plan for Oldest Repository query: https://explain.depesz.com/s/dUTR


santhanakarthikeyan:
I've also implemented the eager loading to fetch all running and failed jobs.

PDegenPortnoy:
This looks good. We're going to want to know the oldest `code_location` but can come up with a better way to identify it

PR #1163, opened 2017-05-26
Fix CommitController#index page performance issue

santhanakarthikeyan:


santhanakarthikeyan:
I ran the below test in my localhost against staging-db for firefox project. 

Before fix:
  Completed 200 OK in 3422ms (Views: 666.7ms | ActiveRecord: 2608.5ms)

After fix: 
  Completed 200 OK in 2172ms (Views: 541.6ms | ActiveRecord: 1269.0ms)

PR #1164, opened 2017-05-29
OTWO-4774 Fix CommitShow Performance Issue

rapbhan:
**Before FIX** (on stage web3 console)
```ruby
irb(main):001:0> Benchmark.ms { SlocMetric.commit_summaries(477942583, 29673307) }
=> 4172.802314162254
```

**After FIX**(on stage web3 console)
```ruby
irb(main):001:0> Benchmark.ms { SlocMetric.commit_summaries(477942583, 29673307) }
=> 288.65329921245575
```


santhanakarthikeyan:
How about rewriting this query into a single query like below

```
SlocMetric.select_summary_attributes
          .select('languages.nice_name as language_name, languages.name as name')
          .joins([[diff: [commit: :fyle]], :language])
          .where(diffs: { commit_id: commit })
          .where('diffs.fyle_id = fyles.id')
          .where(sloc_set_id: AnalysisSlocSet.for_analysis(analysis_id).select(:sloc_set_id))
          .ignored_files(analysis_id)
          .group([:language_id, 'languages.nice_name', 'languages.name'])
```

PR #1166, opened 2017-05-31
Fix test case failures

santhanakarthikeyan:


santhanakarthikeyan:
retest this please

santhanakarthikeyan:
Finally i could see the green build in jenkins :).

test/models/commit_flag/back_to_life_test.rb
```diff
@@ -12,7 +12,6 @@ class CommitFlag::BackToLifeTest < ActiveSupport::TestCase
     btl = CommitFlag::BackToLife.find(cf.id)
     btl.time_elapsed = 123.0
     btl.save!
-    btl.reload
     btl.time_elapsed.must_equal 123.0
   end
```
PDegenPortnoy: Wait, does this not work if we do `btl.reload`? That seems odd that we have to have the object in the same state.

PR #1171, opened 2017-06-01
OTWO-4789 Change to project model to delete associated enlistments 

msk999:
and add script to fix existing data

https://jira.dc1.lan/browse/OTWO-4789


santhanakarthikeyan:
retest this please

script/sql/update_deleted_enlistments_for_deleted_projects.rb
```diff
@@ -0,0 +1,14 @@
...
+
+require_relative '../config/environment'
+
+# update enlistments deleted field for all deleted projects
+ActiveRecord::Base.connection.execute(
+ "update enlistments set deleted = true
+   from enlistments e
+     inner join projects p on e.project_id = p.id
+     where p.deleted and NOT e.deleted"
+)
```
PDegenPortnoy: We may need to do more because this will bypass the Edit mechanism.  Something like:

```ruby
editor = Account.find_by_login('ohloh_slave')
Project.find_in_batches(deleted: true) do |projects|
  projects.each do |project|
    project.enlistments.each do |enlistment|
      enlistment.create_edit.undo!(editor) if enlistment.create_edit.allow_undo?
    end
  end
end
```

msk999:
@blackducksoftware/ohloh-developers 

Pull request changes have been made and tested.  Please review and comment.

script/update_deleted_enlistments_for_deleted_projects.rb
```diff
@@ -3,18 +3,23 @@
 raise 'RAILS_ENV is undefined' unless ENV['RAILS_ENV']
 
 require_relative '../config/environment'
+require 'logger'
+$LOG = Logger.new('update_deleted_enlistments_log_file.log')
```
PDegenPortnoy: This is fine for this script, since it will only be loaded when run, but the `$VARIABLE` syntax is potentially dangerous because it defines a global variable, which, if previously defined would alter that definition. Simply using an initial capital letter, such as "Log" would create a constant and that would generate a warning should something try to reassign a value to it. 

One is well served by creating a class just for the purpose of the script, for example `class UpdateDeletedEnlistmentsForDeletedProjects` and then take advantage of instance variables such as `@log`.

A nice and short example is: https://github.com/blackducksoftware/ohloh-ui/blob/master/script/fix_updated_at_for_topics

It's up to you whether or not you want to refactor the code or not.

PDegenPortnoy:
This looks very nice. 

At this point, we would typically consider squashing all the commits by rebasing the branch. 

+1 to merge

msk999:
Merged manually

app/models/project.rb
```diff
@@ -119,6 +120,13 @@ def remove_people
     Person.where(project_id: id).destroy_all
   end
 
+  def remove_enlistments
+    enlistments.each do |enlistment|
+        enlistment.deleted = true
+        enlistment.save
+    end
+  end
```
PDegenPortnoy: This should do the same thing as the script.

```ruby
# @editor is the the current user
enlistments.each do |enlistment|
  enlistment.create_edit.undo!(@editor) if enlistment.create_edit.allow_undo?
end
```

PR #1172, opened 2017-06-01
Removed database_cleaner gem

santhanakarthikeyan:


PDegenPortnoy:
This looks sufficiently straightforward. When did we add the `database_cleaner` gem?

santhanakarthikeyan:
I've added as part of #1166

PR #1174, opened 2017-06-02
Moved FDW update credentails from seed.rb  to new rake task

santhanakarthikeyan:


santhanakarthikeyan:
Note: Once after merging this branch, we should change jenkins to use this rake task.

lib/tasks/fix_fdw_credentials.rake
```diff
@@ -0,0 +1,35 @@
...
+  second_base_execute(alter_fdw_server_query(server, config))
+  second_base_execute(alter_fdw_user_query(server, config))
+end
+
+def alter_fdw_server_query(server_name, config)
+  "ALTER SERVER #{server_name} OPTIONS(set host '#{config['host']}', set dbname '#{config['database']}', set port '#{config['port']}')"
+end
+
+def alter_fdw_user_query(server_name, config)
+  "ALTER USER MAPPING FOR #{config['username']} SERVER #{server_name} OPTIONS(SET password '#{config['password']}')"
```
Priya5: Will this change user mapping if we are using this rake task on the staging database which is restored from production database because we have different users in staging and production database ?

santhanakarthikeyan: Yes i agree it won't work. I'll change the code to drop and re-create user mapping, that is more safer way to alter the user mapping.

PDegenPortnoy: Hey, can't we just `CREATE USER MAPPING...`? In order to drop the user mapping, you have to know what that is and we probably shouldn't hard code it.  However, all we need is the user mapping to match the current environment

Priya5: We can use `DROP USER MAPPING IF EXISTS for username server ...` so that we will only drop the mapping if one exists with matching current environment configurations and then recreate it with required options:
`CREATE USER MAPPING for openhub_user server fis OPTIONS(user 'fisbot_user', password 'XXXXXXX')`

PDegenPortnoy: True, we saw that we could use the "IF EXISTS", but since, in the test environment, we will have created a new database for the localhost server and the environment variables won't be referring to the server in the structure.sql file, it seems that it won't ever be possible to 'DROP USER MAPPING' because we have the configurations for the current (i.e. "test") database and not the source (i.e. "development" from the "rake db:structure:dump")

PDegenPortnoy:
This may not work at all  for the test environment.  The Test framework immediately loads and starts to process the factory file, and we never get to the commands in the .rake file

```sh
╭─pdegenportnoy@pcm-pdegenpor-1 ~/src/ohloh-ui ‹2.2.5› ‹add_fix_fdw_rake_task*›
╰─$ RAILS_ENV=test rake fix_fdw_credentials
rake aborted!
ActiveRecord::StatementInvalid: PG::SqlclientUnableToEstablishSqlconnection: ERROR:  could not connect to server "fis"
DETAIL:  FATAL:  database "fis_development" does not exist
: SELECT  "slaves".* FROM "slaves" WHERE "slaves"."hostname" = $1  ORDER BY "slaves"."id" ASC LIMIT 1
/Users/pdegenportnoy/src/ohloh-ui/test/factories/jobs.rb:3:in `block (2 levels) in <top (required)>'
/Users/pdegenportnoy/src/ohloh-ui/test/factories/jobs.rb:2:in `block in <top (required)>'
/Users/pdegenportnoy/src/ohloh-ui/test/factories/jobs.rb:1:in `<top (required)>'
/Users/pdegenportnoy/src/ohloh-ui/config/environment.rb:5:in `<top (required)>'
PG::SqlclientUnableToEstablishSqlconnection: ERROR:  could not connect to server "fis"
DETAIL:  FATAL:  database "fis_development" does not exist
/Users/pdegenportnoy/src/ohloh-ui/test/factories/jobs.rb:3:in `block (2 levels) in <top (required)>'
/Users/pdegenportnoy/src/ohloh-ui/test/factories/jobs.rb:2:in `block in <top (required)>'
/Users/pdegenportnoy/src/ohloh-ui/test/factories/jobs.rb:1:in `<top (required)>'
/Users/pdegenportnoy/src/ohloh-ui/config/environment.rb:5:in `<top (required)>'
Tasks: TOP => fix_fdw_credentials => environment
(See full trace by running task with --trace)
```

lib/tasks/fix_fdw_credentials.rake
santhanakarthikeyan: We should first create development database followed by test, by this way we won't face any issues. 

```
rake db:create
rake db:structure:load
rake db:migrate
rake fix_fdw_credentials

RAILS_ENV=test rake db:create
RAILS_ENV=test rake db:structure:load
RAILS_ENV=test rake db:migrate
RAILS_ENV=test rake fix_fdw_credentials
```

PDegenPortnoy: Do the folks at Imaginea have a local  copy of the development database?  If so, what data is in it?  For Ohloh-UI, it is no longer out of the question to have a local development DB with it being "only" 25 GB.  But we still need the FIS DB and that is still around 1.5 TB. Not an option.

Therefore we in Burlington use the development database on the oh-db01.dc1.lan server.

Therefore, we are assuming that the development (and staging) databases are created by the weekly restore process and we will only ever have test databases for Ohloh and FIS locally.

And therefore, we don't need to create the development database.

Finally, I'm not sure this approach changes anything. The issue seems to be that, in the test environment, the test framework immediately load and processes factory files before the rake task is invoked.

It's almost as if we need to modify values in the structure file before `rake db:structure:load`

santhanakarthikeyan: No @PDegenPortnoy, we don't have a local development database. We are using stagingDB for development. 

In the above steps, i mentioned to create development database in local server because when running rake task for test environment, it fails due to non existence of development database. Therefore, i mentioned to create development database first, followed by test database. By this approach we can run rake task without any issues. Later, we can change the **.env** file to point to stagingDB.

When writing this, one basic thing strike me. Why rake task process the factory files?. Now, i  got the root cause of this problem. 

**test/factories/jobs.rb**
```ruby
FactoryGirl.define do
  2   factory :job do
  3     slave Slave.where(hostname: Socket.gethostname).first_or_create!
  4     association :code_location
  5     status Job::STATUS_SCHEDULED
  6   end
  7  end
```
In the above code at line number 3. We are creating *slave* object on the fly instead of using association. Therefore, while loading this factory file, it throws an error. 

I fixed this problem. Now, we no need to create development database in localhost for running test case:)

### Steps to run test case
```ruby
RAILS_ENV=test rake db:create
RAILS_ENV=test rake db:structure:load
RAILS_ENV=test rake db:migrate
RAILS_ENV=test rake fix_fdw_credentials
```

PR #1177, opened 2017-06-07
Add template0 onto database.yml file

drubio1989:


config/database.yml
```diff
@@ -8,6 +8,7 @@ default: &default
   username: <%= ENV["DB_USERNAME"] %>
   password: <%= ENV["DB_PASSWORD"] %>
   database: <%= ENV["DB_NAME"] %>
+  template: template0
```
rapbhan: How about adding only for `development` and `test` mode?

drubio1989: Sure. That should achieve the same result that I want.

PR #1178, opened 2017-06-07
OTWO-4804: Fix UI for project vulnerability report

SouravD18:
- Add vulnerability percentage info
- Fix bar overextending

### DIFF
--- a/app/helpers/project_vulnerability_reports_helper.rb
+++ b/app/helpers/project_vulnerability_reports_helper.rb
@@ -1,15 +1,18 @@
 module ProjectVulnerabilityReportsHelper
   def pointer(pvr, pvs = false)
-    extra_css = ''
     if pvs
       score = pvr.vulnerability_score.to_f
-      percentage = ((score / max_score) * 84).round
+      percentage = ((score / max_score) * 100)
     else
-      score = pvr.security_score.to_f
-      percentage = (score * 84) / 100
-      extra_css = ';width: 32px;' if score >= 100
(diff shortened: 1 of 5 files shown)


SouravD18:
Issue: [https://jira.dc1.lan/browse/OTWO-4804](url)

SouravD18:
Problem & Fix: We used magic numbers to stop the bars overextending. But it creates some visual inconsistency with the data. I fixed it using the bar relative position starting from right instead left. Also added percentage with the score for vulnerability calculation.

drubio1989:
Looks good to me. If the tests clear all green, move the ticket to QA.

SouravD18:
After QA review: added the "float: right" for second bar. The first bar already has "float: right" in it. This fixes the QA issue.

app/helpers/project_vulnerability_reports_helper.rb
```diff
@@ -1,15 +1,18 @@
 module ProjectVulnerabilityReportsHelper
   def pointer(pvr, pvs = false)
-    extra_css = ''
     if pvs
       score = pvr.vulnerability_score.to_f
-      percentage = ((score / max_score) * 84).round
```
rapbhan: I am not sure / recollect why we had this "84"

SouravD18: I think it's a magic number CSS fix. When I tried with 100, (actual percentage), the bar actually overextends. So we used an extra css variable so that whenever the percentage is around 100, we add the extra css to reach the end.

PR #1187, opened 2017-06-16
Otwo 4790 - Implement Job Scheduler changes - migrations

msk999:
There is a csv file attached to the jira ticket that needs to be placed into the tmp directory before running the migrations

The migrations will remove the default on the code_location.status field.  This will allow the status be undefined, it the code location is created via openhub.  The only Active status code_locations are those that have been created via the api and are of interest to the kb team.

The second migration will load the csv file into a temporary table and reset any Active code_locations that are not in the csv file to nil.


db/migrate/20170609195100_remove_codelocation_status_default.rb
```diff
@@ -0,0 +1,9 @@
+class RemoveCodelocationStatusDefault < ActiveRecord::Migration
+  def self.up
+    change_column_default :code_locations, :status, nil
```
santhanakarthikeyan: @msk999, please do let us the reason for changing **code_locations#status** to nil.

db/migrate/20170616152705_set_codelocation_status_to_nil.rb
```diff
@@ -0,0 +1,93 @@
+class SetCodelocationStatusToNil < ActiveRecord::Migration
```
santhanakarthikeyan: We used to create a migration file only to modify the schema. As this file deals with updating the existing values in the table, we could move this file under scripts folder.

db/migrate/20170609195100_remove_codelocation_status_default.rb
santhanakarthikeyan: I apologize, I didn't see the description. IMHO, i would prefer setting a new value to status field instead of nil.

msk999: I could have the status field default to 0 instead of nil

db/migrate/20170616152705_set_codelocation_status_to_nil.rb
msk999: @PDegenPortnoy had asked to have this called in a migration to allow for the ability to rollback the changes if necessary.  @PDegenPortnoy is this what you expected?

db/structure.sql
```diff
@@ -10261,6 +10261,10 @@ INSERT INTO schema_migrations (version) VALUES ('20170323130035');
 
 INSERT INTO schema_migrations (version) VALUES ('20170411054438');
 
+INSERT INTO schema_migrations (version) VALUES ('20170609195100');
+
+INSERT INTO schema_migrations (version) VALUES ('20170616152705');
+
```
Priya5: fisbot/db/secondbase/structure.sql should also be replaced with this structure.sql

db/secondbase/structure.sql
```diff
@@ -1035,23 +1044,153 @@ CREATE FUNCTION vulnerabilities_id_seq_view() RETURNS integer
...
--- Name: USER MAPPING ohloh_user SERVER openhub; Type: USER MAPPING; Schema: -; Owner: -
+-- Name: USER MAPPING fisbot_app_stage SERVER ohloh; Type: USER MAPPING; Schema: -; Owner: -
 --
 
-CREATE USER MAPPING FOR ohloh_user SERVER openhub OPTIONS (
-    password 'password',
-    "user" 'ohloh_user'
+CREATE USER MAPPING FOR fisbot_app_stage SERVER ohloh OPTIONS (
+    password 'Qvnp8Q3mih8IJfMv',
+    "user" 'ohloh_ui_app_stage'
```
santhanakarthikeyan: @msk999 we should not expose these credentials, therefore, please do run the migration against your local development database. I've explained to @drubio1989 probably he could brainstorm you along those lines.

msk999: @santhanakarthikeyan I was wondering why we need to include the structure.sql files at all?  Shouldn't running the migrations create these files anyway?

db/structure.sql
msk999: fixed

PR #1201, opened 2017-07-18
OTWO-4826 Design new possible badges

Priya5:


PDegenPortnoy:
This is not a change set that we are going to merge into master. This was a SPIKE to research options in badge design for the GitHub Integration Project.

PR #1210, opened 2017-08-10
OTWO-4875 Fix demographic charts issue

jebingeosil:


PDegenPortnoy:
After QA review, let's get this into production

PR #1212, opened 2017-08-15
OTWO-4856 Fix memory hog when fetching unclaimed people

notalex:


app/controllers/unclaimed_controller.rb
```diff
@@ -34,4 +34,21 @@ def preload_projects_from_positions
...
+
+  def unclaimed_people(query, find_by, per_page = nil)
+    if query && Person.find_by_name_or_email(q: query).size > OBJECT_MEMORY_CAP
+      unclaimed_people_with_limit(query, find_by)
+    else
+      Person.find_unclaimed(q: query, find_by: find_by, per_page: per_page)
+    end
+  end
+
+  # NOTE: Since this approach avoids the *includes*, it takes 3x DB time. However this prevents memory hog.
```
PDegenPortnoy: And we really only expect this to be invoke when the anonymous user account is included in the query results. Is that correct?

PDegenPortnoy:
These are the reasons for the failing build:

```
test/support/unclaimed_controller_test.rb:2:3: C: Use module_function instead of extend self.
  extend self
  ^^^^^^^^^^^
test/support/unclaimed_controller_test.rb:3:1: C: Trailing whitespace detected.

```

app/controllers/unclaimed_controller.rb
notalex: Yes. Anonymous is the only *name* with over 20k *people* associations.

PR #1213, opened 2017-08-21
OTWO-4856 Fix committers#index memory hog

notalex:
This fixes the case where a high number of people objects cause memory hog in the default index results.


PDegenPortnoy:
@notalex, this looks good to merge

PR #1215, opened 2017-08-24
OTWO-4923: Use dynamic evaluation of time intervals in Ohloh UI ActiveAdmin

Priya5:


PDegenPortnoy:
Very nice solution  🥇 

Looks good to merge

PR #1226, opened 2017-09-07
OTWO-4942: Update fix_fdw_credential task to target ohloh server and …

drubio1989:
…not openhub


rapbhan:
There's is already existing PR https://github.com/blackducksoftware/ohloh-ui/pull/1225 which addresses the same issue.

PR #1232, opened 2017-09-12
Fix Rubocop Offence

rapbhan:


rapbhan:
Merging without PR review as it a fix for rubocop offence

PR #1245, opened 2017-10-06
OTWO-4973 Use modified oink gem to log request url

notalex:


PDegenPortnoy:
This is good. However, will the change to the oink them get a log entry recorded if the process never returns?

notalex:
This does not do that yet. The oink middleware writes to the log file as soon as it is called. I have tried adding oink to the top of the middleware stack, placing a debugger in controller action, but the logs still get written after the request completes.

notalex:
The build failure is due to test coverage. The test coverage is fixed in #1244.

PR #1247, opened 2017-10-11
OTWO-4975 Add gem to kill & log slow requests

notalex:


notalex:
Build is failing due to nokogiri vulnerability and a test failure in master. The test failure in current master is fixed in #1244

PR #1252, opened 2017-10-18
OTWO 4977 Save Github login during github verification

notalex:


PDegenPortnoy:
This looks good.  Merging to Master so we can see it work in production.

PR #1256, opened 2017-10-24
OTWO 4978: Log bloated requests and kill them

notalex:


msk999:
lgtm - confirmed qa steps in jira ticket as well.

PR #1258, opened 2017-10-31
Update nokogiri gem with security patch

alebruck:


notalex:
Thank you @alebruck for this patch.

PR #1260, opened 2017-10-31
Rename FactoryGirl to FactoryBot

alebruck:


notalex:
The build is failing due to the following lint failure.
```
test/models/analysis/commit_volume_test.rb:10:44: C: Align the elements of a hash literal if they span more than one line.
                                           position: 0)
                                           ^^^^^^^^^^^
```

If we move the text **position: 0)** left by one space, it should fix this failure.

alebruck:
My bad, I'll update the PR soon.

PR #1264, opened 2017-11-08
OTWO-4995: Replace CTA with CoPilot

PDegenPortnoy:
We are done with the 2017 Survey and need to replace the Call to Action.
However, when testing the change on Safari, the author noticed that the
search bar was not properly positioned.  Therefore, the home page
Billboard was reworked using the Bootstrap grid to ensure the page
displays the same.

Also note that the header banner has been removed from all non-home
pages.


PDegenPortnoy:
Failing test have to do with 0.08% shortage in code coverage and a problem with the GitHub API for reporting results.  Merging, retesting, and deploying 'cause we need to get the Survey CTA down

PR #1269, opened 2017-11-24
OTWO-5013: Create new enlistment with same repo but different branch

santhanakarthikeyan:


app/controllers/concerns/enlistment_filters.rb
```diff
@@ -64,7 +67,36 @@ def validate_project
...
+  def valid_code_location?
+    @code_location = build_code_location
+    @repository = @code_location.repository
+    return render :new, status: :unprocessable_entity unless @code_location.valid?
+  end
+
+  def code_location_exist?
+    code_location = CodeLocation.find_existing(@repository.url, @code_location.module_branch_name)
+    return unless code_location
+    update_repo_username_and_password(code_location)
```
PDegenPortnoy: If we find an existing Code Location, why is it important to update the username and password? Were we doing this somewhere else or is this needed functionality that was not previously present?

notalex: We were doing it previously in the controller (here|https://github.com/blackducksoftware/ohloh-ui/blob/master/app/controllers/enlistments_controller.rb#L85). This handles the scenario where a code location is present but its repository's username and password have changed.

PDegenPortnoy: That's the use case I had assumed.  Very good.

PDegenPortnoy:
We have conflicts that prevent the branch from being merged and the failure in `test/models/reverification_tracker_test.rb:324`

santhanakarthikeyan:
Fixed merge conflicts and Rubocop issues. Merging this PR.

PR #1272, opened 2017-11-28
Revert change made in c1f243c5

notalex:
We want to display the disabled.html page when one tries to access adisabled account's page.

### DIFF
--- a/app/controllers/concerns/redirect_if_disabled.rb
+++ b/app/controllers/concerns/redirect_if_disabled.rb
@@ -1,8 +1,7 @@
 module RedirectIfDisabled
   def redirect_if_disabled
-    account = current_user
-    return unless account && account.access.disabled?
-    request.env[:clearance].sign_out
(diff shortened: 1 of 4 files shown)


app/controllers/concerns/redirect_if_disabled.rb
```diff
@@ -1,8 +1,7 @@
 module RedirectIfDisabled
   def redirect_if_disabled
-    account = current_user
-    return unless account && account.access.disabled?
-    request.env[:clearance].sign_out
-    redirect_to disabled_account_url(account)
+    return unless @account && @account.access.disabled?
+    request.env[:clearance].sign_out if @account.id == current_user.id
+    redirect_to disabled_account_url(@account)
```
PDegenPortnoy: This looks good -- signing out the user if the current user is disabled

PDegenPortnoy:
+1 to merge

PR #1276, opened 2017-12-08
OTWO-5039 Restore login with email

PDegenPortnoy:
With the migration to Clearance, we lost the ability to login with an
email address.  This restores that feature.

There was also a regression failure where a disabled account was being
redirected to the "authenticate new" path. This was due to the
"!current_user.access.disabled?" in the
ApplicationController#redirect_for_spammer_verification method.  When a
user is disabled, then they'll get caught by a later before_action
filter, so this check isn't needed here.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -255,8 +255,7 @@ class ApplicationController < ActionController::Base
   end
 
   def redirect_for_spammer_verification
-    #return if current_user && !current_user.access.disabled? && current_user.access.mobile_or_oauth_verified?
-    return if current_user  && current_user.access.mobile_or_oauth_verified?
+    return if current_user && current_user.access.mobile_or_oauth_verified?
     redirect_to new_authentication_path
   end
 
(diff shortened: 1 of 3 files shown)


rapbhan:
retest this please

PDegenPortnoy:
@rapbhan I still get a test failure:

```
  1) Failure:
AccountsController::edit#test_0007_must logout spammer trying to edit or update [/Users/pdegenportnoy/src/ohloh-ui/test/controllers/accounts_controller_test.rb:221]:
Expected response to be a redirect to <http://test.host/accounts/login-1988/disabled> but was a redirect to <http://test.host/authentications/new>.
Expected "http://test.host/accounts/login-1988/disabled" to be === "http://test.host/authentications/new".
```

PDegenPortnoy:
I am using the latest from the branch.  Just confirmed that I have your latest changes.

rapbhan:
Yes, I did had the same test failed on my local and after a couple of iterations, merging the `master` branch seems to fix the issue and it's all clean now.

PR #1277, opened 2017-12-09
[OTWO-5030] Added RegistrationKeys table and model

santhanakarthikeyan:


PDegenPortnoy:
Thank for thinking of this, but Ohloh UI really doesn't need to know about eFISbot client registration keys. I'm perfectly content with these keys being administered only through eFISbot.

santhanakarthikeyan:
Thanks @PDegenPortnoy for the clarification. Moving ActiveAdmin pages to eFISbot, hence closing this PR

PR #1279, opened 2017-12-11
OTWO-4970 Signup/signin using Github oauth

notalex:
Review per commit to prevent brain damage!


Priya5:
LGTM

notalex:
The build fails due to test failure for positions_factories_controller_test.rb:53. This test passes locally. It seems like an intermittent failure not caused by changes in this PR.

PR #1287, opened 2017-12-24
OTWO-5064 Restrict github verification 

notalex:


app/controllers/authentications_controller.rb
```diff
@@ -96,4 +97,14 @@ def redirect_if_current_user_verified
     return if current_user.nil?
     redirect_to root_path if current_user.access.mobile_or_oauth_verified?
   end
+
+  def redirect_invalid_github_account
+    return if github_api.created_at < 1.month.ago && github_api.repository_has_language?
```
PDegenPortnoy: Don't we want this check to be for no repositories?

`return if github_api.created_at < 1.month.ago && !github_api.repository_has_language?`

config/locales/en.yml
```diff
@@ -410,6 +410,8 @@ en:
       mobile_verification_hint: Please verify your account using either your mobile number or Github account. OpenHub will not store your mobile number. Only your Github email and login will be kept by OpenHub.
     create:
       error: There seems to be an issue with the mobile verification. Please try again later.
+    github_callback:
+      invalid_github_account: This github account does not fulfill our signup requirements.
```
PDegenPortnoy: Let's change this (just wordsmithing): "This github account is insufficient for signing up at the Open Hub."

That's a little less precise.  That said, this is a public repository, so some clever person could check the code.  :-)

test/controllers/authentications_controller_test.rb
```diff
@@ -161,33 +197,51 @@
...
+      it 'wont sign in a non github verified user which fails github restrictions' do
+        github_account_stub.stubs(:created_at).returns(Time.current)
+        @controller.stubs(:github_api).returns(github_account_stub)
+        account.verifications.delete_all
+        FirebaseVerification.any_instance.stubs(:generate_token)
+        create(:firebase_verification, account: account)
+
+        get :github_callback, code: Faker::Lorem.word
+
+        request.env[:clearance].current_user.must_be_nil
```
PDegenPortnoy: Have I missed the test case where the account is > 1.month, but there are no repositories with recognizable code?

app/controllers/authentications_controller.rb
notalex: The `created_at < 1.month.ago` comparision can be a little confusing due to how datetime comparision works. If I didn't have tests, I wouldn't have been confident about this code.

This line can be translated as:
```ruby
return if account_is_older_than_a_month? && account_has_a_repository_with_language?
redirect_to path, message: 'This account does not fulfill our requirements'
```
Does it feel correct now?

config/locales/en.yml
notalex: Fixed.

test/controllers/authentications_controller_test.rb
notalex: Yes, it is at line 104. 
https://github.com/blackducksoftware/ohloh-ui/blob/562eb5e8fd7760ed34a52ee6a5d911af478ed99d/test/controllers/authentications_controller_test.rb#L104

This uses a response recorded from authenticating a github user alex-bds.

test/controllers/authentications_controller_test.rb
```diff
@@ -161,33 +197,51 @@
...
+      end
+
+      it 'must sign in a verified user regardless of github restrictions' do
+        github_account_stub.stubs(:created_at).returns(Time.current)
+        @controller.stubs(:github_api).returns(github_account_stub)
+
+        get :github_callback, code: Faker::Lorem.word
+
+        request.env[:clearance].current_user.id.must_equal account.id
+      end
```
notalex: This test explains that once a user is github verified by us, we would sign them in regardless of their github activity. We do this to ensure a snappy sign in.

test/controllers/authentications_controller_test.rb
```diff
@@ -161,33 +197,51 @@
...
         account.reload
         must_redirect_to account
         request.env[:clearance].current_user.id.must_equal account.id
-        account.github_verification.token.must_equal github_stub.access_token
+        account.github_verification.token.must_equal github_account_stub.access_token
+      end
+
+      it 'must sign in a verified user regardless of github restrictions' do
+        github_account_stub.stubs(:created_at).returns(Time.current)
+        @controller.stubs(:github_api).returns(github_account_stub)
```
notalex: This test explains that once a user is github verified by us, we would sign them in regardless of their github activity. We do this to ensure a snappy sign in.

app/controllers/authentications_controller.rb
PDegenPortnoy: Yes, these are both positive tests: the account must be older than 1 month and there is a public repository with recognizable code.  Thanks for the explanation,

test/controllers/authentications_controller_test.rb
PDegenPortnoy: Thank you

PR #1290, opened 2018-01-09
Added the NewRelic License via ENV

rapbhan:


karthikeyanBD:
We should add this key in openhub-config other than that it LGTM.

rapbhan:
@karthikeyanBD Yes, it has already been added

PDegenPortnoy:
Good to merge when we have the new key in the Config repository

PR #1292, opened 2018-01-12
OTWO-5065 Script to mark accounts with insuffient github verification criteria as spammer

Priya5:


PDegenPortnoy:
This looks good, @Priya5.  Let's merge to master and run it in staging and then in producution

PR #1294, opened 2018-01-23
OTWO-5033 Update Enlistments to use CodeLocation Subscription

rapbhan:


notalex:
:+1:

notalex:
It would be helpful to have the CodeLocationSubscription inherit from a FisbotApi class. There would be other models communicating with fisbot. A common parent would ease adding new changes.

notalex:
:+1:

PR #1298, opened 2018-01-31
OTWO-5033 Fixed the Load Path

rapbhan:
I have moved the classes from `app/lib/` to `lib/`. The reason being otherwise we have to load the path for the class(es) as below and so to avoid it (as stated below)

```ruby
class CodeLocationSubscription < Fisbot::FisbotApi
end
```

and similarly in the parent class `FisbotApi`

```ruby
class Fisbot::FisbotApi
end
```


PDegenPortnoy:
I don't have any objection to this change per se. Was the `app/lib` a directory created and expected by Rails?

notalex:
The `app/lib` was a directory created and custom loaded by us. It is intended to contain supporting libraries like **acts_as_editable**, **filter_by** etc. In our current structure, **lib** contains scripts like some regex patterns, rake tasks, OpenhubSecurity, `strip_attributes` method. These are small scripts compared to complete libraries in **app/lib**.

PR #1301, opened 2018-02-14
OTWO-5095 Implemented ProjectJobs API Listing

rapbhan:


app/controllers/oh_admin/jobs_controller.rb
```diff
@@ -0,0 +1,17 @@
+class OhAdmin::JobsController < ApplicationController
+  before_action :admin_session_required
+  before_action :find_project
+  layout 'admin'
+  helper JobApiHelper
+
+  def index
+    @response = JSON.parse(ApiJob.new(@project.id, params[:page]).get)
```
karthikeyanBD: How about renaming the `ApiJob` class to `JobApi`. As per rails convention, get method is to get a single object, not a collection, can we change that to `fetch`/`where`. 

It would be great if we can move the `get` method to a class method.

app/helpers/application_helper.rb
```diff
@@ -127,6 +127,11 @@ def needs_login
     logged_in? ? '' : 'needs_login'
   end
 
+  def api_pagination(response)
+    numbers = (1..response['total_entries']).to_a
+    will_paginate(numbers.paginate(page: response['current_page'], per_page: response['per_page']))
```
karthikeyanBD: This one looks like a hacky solution. As per the rails convention, it seems, we should show `first`, `previous`, `next` and `last` links https://github.com/davidcelis/api-pagination

app/helpers/job_api_helper.rb
```diff
@@ -0,0 +1,54 @@
...
+    time_ago_in_words(updated_at.to_datetime) if updated_at
+  end
+
+  def status_tag(status)
+    case status
+    when 0 then ['scheduled', 'label-warning']
+    when 1 then ['running',   'label-primary']
+    when 2 then ['paused',    'label-warning']
+    when 3 then ['failed',    'label-danger']
+    when 5 then ['completed', 'label-success']
```
karthikeyanBD: We could move these status constants to Job model.

app/admin/jobs_admin.rb
```diff
@@ -115,13 +68,15 @@
...
-        project_jobs
       else
         super
       end
     end
 
+    def index
+      redirect_to oh_admin_project_jobs_path if params[:project_id]
+    end
+
```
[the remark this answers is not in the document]
rapbhan: This was kept for the convenience thing :) ..The admins would have the history of project jobs and thought it would be handy and also if they land from other links (if any) would still be redirected. I am OK if its needs to be removed

app/assets/stylesheets/admin.css
```diff
@@ -0,0 +1,6 @@
+/*
+*= require ./admin/bootstrap.min
+*= require ./admin/font-awesome.min
+*= require ./admin/AdminLTE.min
+*= require ./admin/skin-blue.min
+*/
```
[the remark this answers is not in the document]
rapbhan: Yes, just checked in staging ...the admin css didn't get loaded. Also, the admin has a separate `layout`. However, re-confirmed it in staging.

PR #1312, opened 2018-04-13
OTWO-5153 Github Forge match must return https url

notalex:
We use the forge match to autopopulate the repository url in check_forge
form. Until now we were populating the git url fetched from github.
However fetch using git protocol sometimes fails. Hence we must use
https by default.


PDegenPortnoy:
Retest this please

PR #1318, opened 2018-06-17
OTWO-5228 Restored the Deleted Enlistment

rapbhan:


app/controllers/concerns/enlistment_filters.rb
```diff
@@ -76,8 +76,33 @@ def build_code_location
...
+    enlistment = get_enlistment
+    enlistment.try(:deleted?) ? true : false
+  end
+
+  def get_code_location_id
+    CodeLocationApi.id_from_url(@code_location.url, @code_location.branch)
+  end
+
+  def get_enlistment
+    Enlistment.where(project_id: @project.id, code_location_id: get_code_location_id).first
```
karthikeyanBD: We should use the Subscription model.

app/controllers/concerns/enlistment_filters.rb
```diff
@@ -76,8 +76,33 @@ def build_code_location
   def project_has_code_location?
     return unless CodeLocationSubscription.code_location_exists?(@project.id, @code_location.url,
                                                                  @code_location.branch, code_location_params[:scm_type])
-
-    flash[:notice] = t('.notice', url: @code_location.url, module_branch_name: @code_location.branch)
+    manage_deleted_enlistment
```
karthikeyanBD: I guess we should move this responsibility to Fisbot Api.

rapbhan:
@karthikeyanBD This is a different scenario ...where the enlistment would have been deleted ...and still the subscription exists  ....so we're attempting to undo the enlistment that was deleted ...(currently we show a error message that repository exists). A ideal scenario.... when the enlistments created the subscription gets created ...and the enlistment is deleted the subscription gets deleted ...that's working fine ...this is the other (edge) case as explained above.

app/controllers/concerns/enlistment_filters.rb
rapbhan: The `code_location_id` we get it via API call from Fisbot

app/controllers/concerns/enlistment_filters.rb
PDegenPortnoy: The FISbot should not know anything about projects or enlistments or analyses, etc. It is only concerned with Code Locations and determining whether or not a Code Location needs to be updated.

app/controllers/concerns/enlistment_filters.rb
PDegenPortnoy: The Subscription model relies upon the FDW, which will be going away entirely so that all communication is through the API. The Ohloh UI application should know about the eFISbot service (the API) and nothing about its internal workings.

app/lib/fisbot/code_location_api.rb
```diff
@@ -0,0 +1,13 @@
+class CodeLocationApi < FisbotApi
+  class << self
+    def resource
+      :code_locations
+    end
+
+    def id_from_url(url, branch)
+      query_params = { url: url, branch: branch }
+      uri = api_access.resource_uri('id_from_url', query_params)
```
PDegenPortnoy: This isn't needed.  The existing eFISbot Code Location API provides querying for the `CodeLocation` with the URL and Branch and the returned data package includes the CodeLocation ID.

See: https://confluence.dc1.lan/display/O2/2016-2017+eFISbot+API+Specification+and+Design#id-2016-2017eFISbotAPISpecificationandDesign-GetsingleCodeLocation

rapbhan:
@PDegenPortnoy Yes, now it's using the existing code location API to get the code location id. Thank you!

app/controllers/concerns/enlistment_filters.rb
```diff
@@ -76,8 +76,33 @@ def build_code_location
...
+    end
+  end
+
+  def enlistment_deleted?
+    enlistment = get_enlistment
+    enlistment.try(:deleted?) ? true : false
+  end
+
+  def get_code_location_id
+    CodeLocationApi.id_from_url(@code_location.url, @code_location.branch)
```
PDegenPortnoy: This needs to be updated to use the CodeLocation GET call with the url and branch and then extract the ID from the return package.

app/lib/fisbot/code_location_api.rb
```diff
@@ -0,0 +1,14 @@
+class CodeLocationApi < FisbotApi
+  class << self
+    def resource
+      :code_locations
+    end
+
+    def id_from_url(url, branch)
```
PDegenPortnoy: This method can be deleted.

test/controllers/enlistments_controller_test.rb
```diff
@@ -241,6 +263,7 @@
 
           assert_no_difference 'Enlistment.count' do
             WebMocker.code_location_exists(true)
+            CodeLocationApi.stubs(:id_from_url).returns('1')
```
PDegenPortnoy: This needs to be updated.

rapbhan:
@PDegenPortnoy Yes, its updated now!

PDegenPortnoy:
There are a number of outdated changes in this. Shall we refresh or delete?

rapbhan:
@PDegenPortnoy The latest master has been merged into this branch and the CI has passed. Good to merge it

PR #1319, opened 2018-06-27
OTWO-5264 Fix the Enlistment Page for Failed Jobs

rapbhan:


app/models/job.rb
```diff
@@ -64,5 +64,9 @@ class << self
     def incomplete_project_job(project_ids)
       where(project_id: project_ids).where.not(status: STATUS_COMPLETED).first
     end
+
+    def incomplete_fis_jobs
+      where.not(type: %w(TarballJob VulnerabilityJob), status: STATUS_COMPLETED)
+    end
```
karthikeyanBD: Can we use `scope` to reduce this in a single line?

rapbhan: I did the scope initially and robocop complained about code complexity in enlistment controller and hence I used a method.

rapbhan: Yes, its updated now using `scope`

app/decorators/code_location_job_progress.rb
```diff
@@ -50,7 +50,8 @@ def no_job
     if sloc_set_code_set_time
       I18n.t 'repositories.job_progress.update_completed', at: time_ago_in_words(sloc_set_code_set_time)
     else
-      incomplete_job = Job.where(code_location_id: @project.enlistments.pluck(:code_location_id)).incomplete.first
+      code_location_id = @project.enlistments.pluck(:code_location_id)
+      incomplete_job = Job.where(code_location_id: code_location_id).incomplete_fis_jobs.first
```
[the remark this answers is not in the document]
rapbhan: The Jobs API has only project jobs (based on project_id and not code_location_id). Probably we can add this as a new feature. But for now, this PR just filters out the jobs that are misleading info about the code_location progress.

PR #1327, opened 2018-10-15
OTWO-5359: Fix recaptcha failure

karthikeyanBD:


PDegenPortnoy:
If adding the reCAPTCHA script to a partial is the solution, how did it work before? Is there another partial that is adding this JavaScript?

PR #1328, opened 2018-10-17
OTWO-5338 Log exceptions for fisbot api DELETE endpoint

notalex:


PDegenPortnoy:
This looks good. Let's get this into production and get the deltas between enlistments and subscriptions closed.

PR #1335, opened 2018-11-14
OTWO-5382 Removed duplicate GNU General Public Licenses and added redirection entries to redirect duplicate license vanity URLS to the original license vanity URL

Priya5:


msk999:
LGTM

PR #1336, opened 2018-11-19
OTWO-5400 Update dependencies to address reported CVEs

notalex:


app/controllers/application_controller.rb
```diff
@@ -292,6 +292,7 @@ def check_project_authorization
 
   def update_last_seen_at_and_ip
     return unless logged_in?
+    # rubocop:disable Rails/SkipsModelValidations # We want to skip validations here.
```
Priya5: Can we disable this check directly by adding it to .rubocop.yml file ?

notalex: We do not want to add a catch-all for this, because we want to catch unwanted validation skips in future.

PR #1339, opened 2018-12-04
OTWO-5416 Moved the Recent Active Accounts to rake task

rapbhan:


rapbhan:
restest this please

app/decorators/home_decorator.rb
```diff
@@ -12,7 +12,9 @@ def most_active_projects
   end
 
   def most_active_contributors
-    Account.recently_active.includes(best_vita: [:name_fact])
+    Rails.cache.fetch('HomeDecorator-recently_active_accounts-cache') do
+      Account.recently_active.includes(best_vita: [:name_fact]) if Rails.env.test?
```
PDegenPortnoy: We should remove the `Rails.env.test` conditional and ensure that the code works with one execution pathway. To do that in test, we can mock `Rails.cache` and respond to the `fetch` method so that, in the test environment, it appears to be retrieving these data from cache.

For now, let's get this into production to address the home page issue that is regularly causing the site to become unresponsive.

PDegenPortnoy:
This is good to merge and get into production. We'll address the test specific code in a new scope of work, but it is urgent to get this improved behavior into production.

PR #1342, opened 2018-12-05
OTWO-5418 Fixed the github login to do case insensitive search by account login

Priya5:
- This PR will also add functionality to check for an account matching with Github secondary email addresses.
- Also avoid multiple click on Sign in with Github.

### DIFF
--- a/app/assets/javascripts/github_auth.js.coffee
+++ b/app/assets/javascripts/github_auth.js.coffee
@@ -3,6 +3,7 @@ class App.GithubAuth
 
   authenticate: ($githubButton) ->
     $githubButton.click ->
+      $(this).attr 'disabled', 'disabled'
       github_url_params = [
         AUTH_URL
         '?client_id=', $(this).data('clientId')
--- a/app/models/github_api.rb
+++ b/app/models/github_api.rb
@@ -35,6 +35,12 @@ class GithubApi
     @access_token = data['access_token'].first
   end
 
+  def secondary_emails
+    return @secondary_emails if @secondary_emails
+    @secondary_emails = fetch_email_responses.select { |hsh| !hsh['primary'] && hsh['verified'] }
+                                             .map { |response| response['email'] }.compact
+  end
+
   private
 
   def request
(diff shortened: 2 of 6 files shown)


app/assets/javascripts/github_auth.js.coffee
```diff
@@ -3,13 +3,15 @@ class App.GithubAuth
...
-      window.location.href = github_url_params.join('')
+      if !$(this).hasClass('clicked')
+        $(this).addClass 'clicked'
+        github_url_params = [
+          AUTH_URL
+          '?client_id=', $(this).data('clientId')
+          '&redirect_uri=', $(this).data('redirectUri')
+          '&scope=', $(this).data('scope')
+        ]
+        window.location.href = github_url_params.join('')
```
notalex: instead of adding a class and a conditional, why not simply disable the button on click?
```js
$(this).attr('disabled', 'disabled')
```

app/models/github_api.rb
```diff
@@ -35,6 +35,12 @@ def access_token
     @access_token = data['access_token'].first
   end
 
+  def secondary_emails
+    return @secondary_emails if @secondary_emails
+    @secondary_emails = fetch_email_responses.select { |hsh| !hsh['primary'] && hsh['verified'] }
+                                             .map { |response| response['email'] }.compact
+  end
```
notalex: If this method gets all the emails, it should be named appropriately. **all_emails** or just **emails**.

app/assets/javascripts/github_auth.js.coffee
Priya5: Sure, I will update the code to disable the link after first click.

app/models/github_api.rb
Priya5: This will return non-primary verified emails. Can we name it as verified_emails ?

notalex: sounds good

notalex: We use this method to find the github_api_account. If it doesn't return the primary github email, then we would end up finding only those users whose email on OH is from one of their secondary Github Emails. Is this what we want?

Priya5: We are checking for account with primary email first. If we dont find the account, then we will look for account matching with secondary Github Emails.

>   def github_api_account
    @account ||= Account.find_by(email: github_api.email)
    @account ||= Account.where(email: github_api.secondary_emails).first
  end

notalex: Thanks for pointing it out. In this case the name secondary_emails makes complete sense.

Priya5:
retest this please

PR #1359, opened 2019-01-28
OTWO-5470 Added rake task to fix broken links with URL redirection

Priya5:


lib/tasks/fix_broken_links_with_url_redirection.rake
```diff
@@ -0,0 +1,69 @@
...
+    http = Net::HTTP.new(uri.host, uri.port)
+    http.read_timeout = 30
+    http.open_timeout = 30
+    http.use_ssl = (uri.scheme == 'https')
+    http.verify_mode = OpenSSL::SSL::VERIFY_NONE if uri.scheme == 'https'
+    http
+  end
+
+  def handle_url_redirection(url, limit = 1)
+    return if url.blank? || @link.url == url || limit > @max_limit
```
PDegenPortnoy: Given that there is a max limit defined during initialization, this line could his that max limit and the `handle_url_redirection` could fail silently, but this seems like a small risk that doesn't warrant additional protections.

PDegenPortnoy:
Looks good. +1 to merge

PR #1361, opened 2019-02-01
OTWO-5145 Used cache to store people total count

Priya5:


PDegenPortnoy:
Let's get this into production

PDegenPortnoy:
NOTE: there are the 3 *new* servers to which code must be deployed until the swap is completed

PR #1377, opened 2019-03-01
OTWO-5502 Fixed 'nil is not a symbol nor a string' error

Priya5:


PDegenPortnoy:
Looks good; please merge and deploy

PR #1394, opened 2019-05-03
OTWO-5562 Fixed the 5 Year Trend Chart Image

rapbhan:


app/models/analysis/query_base.rb
```diff
@@ -85,4 +84,16 @@ def within_date
...
+
+  def contributor_monthly_commits_query
+    sql = <<-SQL
+      select to_char(date(C.time),'MON,YYYY') as month, count (*) as count
+      FROM  commits C INNER JOIN code_sets CS ON C.code_set_id = CS.id
+      INNER JOIN sloc_sets SS ON SS.code_set_id = CS.id INNER JOIN analysis_sloc_sets ASS ON ASS.sloc_set_id = SS.id
+      INNER JOIN analysis_aliases K on K.commit_name_id = C.name_id
+      WHERE ASS.analysis_id = #{@analysis.id} AND K.analysis_id = #{@analysis.id} AND C.position <= ASS.as_of AND K.preferred_name_id = #{@name_id}
+      group by month
+    SQL
```
PDegenPortnoy: I have a serious concern since `commits`, `code_sets`, `sloc_sets`, `analysis_sloc_sets`, and `analysis_aliases` are all foreign tables that are in the FIS DB. When we sever the foreign data wrapper between the two databases, these foreign tables go away. 

Commits will be available through the new FIS API Commits end point, but performing this type of query in order to generate the 5 Year Trend Chart via API will clearly not be successful.  Therefore, we should consider extending the VitaJob so that it generates the charts for each user. Are we caching these images? Do we have an idea about how many images we generate and cache?

Let's get the build to pass and get this into production, but we're going to need to address this with FODS 3.0

rapbhan: Yes, we do cache these charts and expires every 4 hours. I am not sure of the count of images generated.

PR #1396, opened 2019-05-15
OTWO-5494 Removed BrakeMan Gem

rapbhan:


PDegenPortnoy:
Should we include an update to the Readme to let users know that the `ci` rake task uses Brakeman and they need to install it outside of the `bundle install`?

notalex:
Not needed anymore.

PR #1397, opened 2019-05-16
OTWO-5591 Refactor vulnerability details select query

Priya5:


PDegenPortnoy:
This looks good. I think the risk of `vulnerability_ids` becoming a *huge* value is low because the domain almost guarantees that there will not be millions and billions of vulnerabilities per release.

PR #1400, opened 2019-05-23
OTWO-5604 - Language page error  deleted accounts

msk999:


msk999:
Issue with deleted accounts will cause an exception on the language show page.  The experienced contributors field on the languages table is not reset after an account is deleted.  This field is only reset when the monthly language cron job is run.

Fix is to test if the account exists in the hash and skip if it is not located

PR #1404, opened 2019-06-30
OTWO-5630: Replace sanitize/escape logic with strip_tags

notalex:
full_sanitizer.sanitize == strip_tags
https://github.com/rails/rails/blob/76a0b10/actionview/lib/action_view/helpers/sanitize_helper.rb#L80


Priya5:
LGTM

PR #1408, opened 2019-07-29
OTWO-5691 Script to cleanup old unused Vulnerabilities & related tables

Priya5:


karthikeyanBD:
Don't merge

lib/tasks/cleanup_vulnerabilities.rake
```diff
@@ -0,0 +1,37 @@
+desc 'Cleanup old unused vulnerabilities data'
+task cleanup_vulnerabilities: :environment do
+  @conn = ActiveRecord::Base.connection
+  Project.where.not(best_project_security_set_id: nil).select(:id, :best_project_security_set_id).find_each do |project|
+    begin
+      pss_ids = ProjectSecuritySet.where(project_id: project.id)
+                                  .where('id < ?', project.best_project_security_set_id).select(:id).pluck(:id)
```
PDegenPortnoy: Isn't thing potentially dangerous when DB IDs wrap around?  We would have old ID's that are larger than the `project.best_project_security_set_id` that are *supposed* to be removed, but won't be with this logic. Is there a `created_at` field we can use?

karthikeyanBD:
I've migrated the Ohloh script for Ohloh-UI 
```
#!/usr/bin/env ruby
ENV['RAILS_ENV'] ||= 'development'
require_relative 'config/environment'

module CleanupVulnerabilities
	def self.run!
		i = 1
		ProjectSecuritySet.select('distinct project_security_sets.id').joins('inner join jobs on jobs.project_id = project_security_sets.project_id left outer join projects on projects.best_project_security_set_id = project_security_sets.id').where("projects.best_project_security_set_id is null and jobs.type = 'VulnerabilityJob' and jobs.status in (3,5)").find_in_batches(batch_size: 100) do |pss_ids|
			puts "iteration ##{i}"
			i = i + 1
			pss_ids = pss_ids.map(&:id)
			release_ids = Release.where(project_security_set_id: pss_ids).ids
			vulnerability_ids = ReleasesVulnerability.where(release_id: release_ids).pluck(:vulnerability_id)

			Vulnerability.where(id: vulnerability_ids).delete_all
			ReleasesVulnerability.where(release_id: release_ids).delete_all
			Release.where(id: release_ids).delete_all
			ProjectSecuritySet.where(id: pss_ids).delete_all
			puts "done"
		end
	end
end

CleanupVulnerabilities.run!
```

PR #1410, opened 2019-08-02
OTWO-5690 Add health route for monitoring

notalex:


PDegenPortnoy:
672 errors, 'ActiveRecord::NoDatabaseError: FATAL:  role "jenkins" does not exist'

notalex:
The **/health** route would resolve if ActiveRecord has a connection to the DB. But there are other fail points. Like failure to connect to secondary database, database being out of space etc. We should list out the cases where the **/health** route should fail.

PDegenPortnoy:
Good points @notalex. It looks like `/health` should fail when the Rails app cannot access primary or secondary DB (the DB Out of Space error will cause the DB service to stop), or when it cannot access Sidekiq or Redis.

Does that seem sufficient?

notalex:
Yes. That is correct. We should be covered. This would return 500 in all failure cases.

PR #1411, opened 2019-08-06
OTWO-5157 Add code_location admin url to enlistments page

notalex:


app/helpers/enlistments_helper.rb
```diff
@@ -20,4 +20,8 @@ def sidekiq_work_in_progress?
     key = Setting.get_project_enlistment_key(@project.id)
     Setting.get_value(key).try(:present?)
   end
+
+  def code_location_admin_url(id)
+    "https://efisbot.blackducksoftware.com/admin/code_locations/#{id}/jobs"
```
Priya5: This is the outdated URL of Efisbot.

notalex: what is the updated url?

PR #1413, opened 2019-08-19
OTWO-5708: Upgrade nokogiri to address vulnerability

notalex:
https://github.com/sparklemotion/nokogiri/issues/1892

We had to downgrade bundler as rails 4.2 depends on bundler <2


Priya5:
LGTM

PR #1423, opened 2019-10-16
OTWO-5755: OH Home Page: Update CSS

bostonpops:
This PR addresses the initial restyling of the OH UI home page, using colors that are addressed in Synopsys Color Standards.  ThIs also includes adding a new logo, using Roboto as the site font, and adding new links in the footer.


app/assets/stylesheets/footer.sass
```diff
@@ -1,53 +1,60 @@
...
-      a
-        color: white
-        padding: 8px 12px
-        &:hover
-          background: #0082C6
-          border-radius: 0
+    margin-top: 20px
+    margin-left: 30px
+    width: 15%
+    color: $secondary-light-purple !important
```
msk999: Should we be using important rule as much.  It's my understanding that this is not a best practice.  Here's a link that talks about when it should be used.  https://css-tricks.com/when-using-important-is-the-right-choice/

app/assets/stylesheets/page.sass
```diff
@@ -1,8 +1,8 @@
 @import "base"
 
 body
-  background-color: #555
-  font-family: Open Sans
+  background-color: white
+  font-family: Roboto
```
msk999: I assume Roboto is the new standard?

app/assets/stylesheets/page.sass
```diff
@@ -14,7 +14,7 @@ body
   padding: 10px 10px 0 10px
 
 header
-  background-color: #242021
+  background-color: white
```
msk999: The color keywords is so much clearer!  I don't see white as one of our defined variables.  Is white always the same white across all browsers?  I see we are replacing different colors with white . (see line 83 below) . Is that intentional?

app/assets/stylesheets/footer.sass
bostonpops: Agreed.  There's lots of `!important` scattered thru all of the code.  After touching up a few pages, I'll have a better feeling on how best to minimize `!important` for common UIs.

app/assets/stylesheets/page.sass
bostonpops: That's correct.

app/assets/stylesheets/page.sass
bostonpops: We decided to use `white` as the background, and perhaps it is a better practice to define something like `$white: #ffffff` rather than rely on the CSS name iteself.

PDegenPortnoy:
This looks really good! Of course, I have some comments :-)

1. Let's remove the underscore under "Black Duck Open Hub". It's pretty standard that clicking on that will bring a user back to the home page
1. For the Twitter link, I think we can do better than the current "twitter.png" image with the OH logo. Too much pixellation. Could we try doing it with text and CSS? If not, please let me know and I'll make a clear PNG file.
1. The "Search Projects..." text. Let's please set that color to #505867, Primary Dark Gray + 20%
1. The "Join Now", "What's New", "Most Popular Projects", "Most Active Projects", and "Most Active Contibutors", let's increase the `font-weight` to 500. Maybe this is a change to all H2
1. It looks like the "Most...." headers should be changed to H2 and loose the `common` style so that they are structurally (for SEO analysis / Semanic Tagging) correct and the proper weight. (Plus that will help with the responsive layout in the futures)
1. The "Join Now" button in the "Join Now" section: Let's get that to be bold. Something is overriding the `font-weight`
1. I didn't specify it, but let's please make the "Synopsys" logo in the footer a link to "https://www.synopsys.com"
1. Finally, either by using 'git rebase -i' or with GitHub, please squash the commits. There probably isn't a need for all these individual commits to be in the history.

bostonpops:
Hi Peter,

So far I've addressed 1,3-7.  I'm working on the css for the "OH" text
(yes, text is a better choice here; thanks for suggesting that).  After
that, you should be able to do a pull and verify all the points above.  You
make another good point on squashing commits (#8 above), interactive
rebasing is definitely a good practice.

Chris

PR #1426, opened 2019-11-04
OTWO-5751 Ohloh UI: Replace gem rubocop by rubocop-rails

bostonpops:
This PR addresses replacing the existing `rubocop` gem with `rubocop-rails`, and correcting any issues found by the new gem.  
Rubocop has moved Rails cops to separate gem 'rubocop-rails'. So, we need to replace 'rubocop' gem with 'rubocop-rails'.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -38,7 +38,7 @@ Lint/ScriptPermission:
   Exclude:
     - script/*
 
-Documentation:
+Style/Documentation:
   Enabled: false
 
 Style/ClassAndModuleChildren:
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -9,10 +9,6 @@ AllCops:
 Rails:
   Enabled: true
 
-Rails/SkipsModelValidations:
-  Exclude:
-    - 'test/**/*.rb'
-
 Rails/UnknownEnv:
   Environments:
     - production
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -9,6 +9,10 @@ AllCops:
 Rails:
   Enabled: true
 
+Rails/SkipsModelValidations:
+  Exclude:
+    - 'test/**/*.rb'
+
 Rails/UnknownEnv:
   Environments:
     - production
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,8 +1,20 @@
+require: rubocop-rails
+
 AllCops:
   TargetRubyVersion: 2.5
   Exclude:
     - vagrant/**/*
     - script/*_temp.rb
+
+Rails:
+  Enabled: true
+
+Rails/UnknownEnv:
+  Environments:
+    - production
+    - development
+    - test
+    - staging
 
 Metrics/LineLength:
   Max: 120
--- a/Gemfile
+++ b/Gemfile
@@ -36,7 +36,7 @@ gem 'rbnacl-libsodium'
 gem 'recaptcha', require: 'recaptcha/rails'
 gem 'redcarpet'
 gem 'redis-rails', '>= 5.0.2'
-gem 'rubocop', require: false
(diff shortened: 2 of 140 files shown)


Priya5:
We need to add the rails cop removed in https://github.com/blackducksoftware/ohloh-ui/commit/139b3adae6b8cf4c128ce46be461be0ea0d16c75#diff-11a0d7906801d9dea0eccb85667ad811

bostonpops:
When I add the Rails cop back w/
```
Rails:
    Enabled: true
```

I now see `Warning: unrecognized cop Rails found in .rubocop.yml` when running `$ rubocop`.  All other behavior seems to be unchahged.

Priya5:
> When I add the Rails cop back w/
> 
> ```
> Rails:
>     Enabled: true
> ```
> 
> I now see `Warning: unrecognized cop Rails found in .rubocop.yml` when running `$ rubocop`. All other behavior seems to be unchahged.

Add `require: rubocop-rails` in the `.rubocop.yml ` file. This will fix the issue.

bostonpops:
@Priya5 Thanks, that worked.  It found 492 offenses!  I'll be posting them shortly...

bostonpops:
Current status of offending cops:
    - `Rails/UnknownEnv:`
        resolved
    - `Rails/HasAndBelongsToMany:`
        resolved
    - `Rails/EnumHash:`
        resolved
    - `Rails/SkipsModelValidations:`
        Rubocop prefers using `update_attributes` as opposed to `update_attribute` or `update_column` as the former will perform validations.
        I turned this cop off on a file-by-file basis with the thinking that w/each violation, we shouldn't blindly be changing code in a situation where an extra validation isn't necessary or shouldn't happen w.r.t performance.
    - `Rails/DynamicFindBy:`
        Rubocop discourages use of the dynamic `find_by_*` methods, and suggest using `find_by` instead.  
        I turned this cop off with the belief that, if necessary, we can add methods to each respective Model for the equivalent functionality provided by the dynamic `find_by_*` methods.
    - `Rails/OutputSafety:`
        Rails discourages use of `.html_safe` and `raw`.  
        In our code, there are many strings with embedded html that require calling `html_safe`.  To address this I've created a single method, `safe_text`,  that accepts a string of chars, calls `html_safe` and returns the resulting string.  `safe_text` is now the only code that calls `html_safe`, and the OutputSafety cop is disabled for this method.
    - `Rails/InverseOf:`
        Rubocop throws this violation when Active Record can't automatically determine the inverse association because of a scope or the options used.
        I turned this cop off on a file-by-file basis thinking that it was too risky to try and address every association that needed updating without breaking the code.
    - `Rails/HasManyOrHasOneDependent:`
        Rubocop wants `has_many` and `has_one` associations to specify a `:dependent` option. It doesn't register an offense if `:through` option was specified.
        I found this cop too risky to resolve without knowing which associations, if necessary, would requrie `delete` vs `destroy` or `delete_all` vs `destory_all`.  The cop is disabled on a file-by-file basis.
    - `Rails/HelperInstanceVariable:`
        Rubocop discourages the use of instance variables in helper classes.
        I resolved this cop up to the point where some helper classes employ memoization where instance vars are a necessary part of the pattern, for these files the cop was turned off.
    - `Rails/LexicallyScopedActionFilter:`
        Rails wants all methods referenced a filter's `only` or `except` options to be present in the implementing class.
        I resolved this copy for simple `ActionController` methods (ie: `create`,`new`,`index`,`show`,`edit`,`update`,`delete`,`destroy`); and turned the cop off, where necessary, for any other methods identified by the cop.

bostonpops:
This pull request will be superseded by a new PR, which will turn off offending cops requiring considerable code changes in the .rubocop,.yml file, and not on a file-by-file basis.

PR #1432, opened 2019-11-14
OTWO 5804 - sort vulnerabilities dropdown by version number

bostonpops:
This PR addresses an issue where the dropdown listing for Vulnerabilities on the `projects#security` page is sorted by respective release date and should be correctly sorted using semantic versioning.  Using a sort by semantic versioning will correctly group all releases listed and not jumble the listing based on the most recent releases; the `projects#security` UI still presents a vulnerability timeline which lists the respective releases for the selected time span.

Refer to https://semver.org for a description on semantic versioning.

This PR converts each respective version string to an array of int for each element in the version string; non-numeric chars, '+', '-', and '_' are translated to their equivalent ASCII values and multiplied by -1 with respect to sorting.  The sorting is done in ascending order, which is then reversed before rendering the dropdown list.  `alpha`, `beta`, and `rc` are reserved terms that have a special sorting order; an `alpha` release precedes a `beta` release which precedes a release candidate (`rc`); the sorting takes this into account and assigns a negative value which sorts these releases with precedence over other 'named' releases, or those which include build information `(.'+...')`.  A few edge cases likely exist with those releases which include build information in that they will be listed immediately _before_ a minor version instead of immediately _after_.

### DIFF
--- a/app/controllers/concerns/vulnerability_filters.rb
+++ b/app/controllers/concerns/vulnerability_filters.rb
@@ -42,6 +42,8 @@ module VulnerabilityFilters
                         @releases.sort_by_release_date
                       end
     @minor_versions = @minor_versions.select_within_years(filter_period_param) if filter_period_param.present?
+    @minor_versions = sort_by_version_number(@minor_versions)
+
     reset_release
   end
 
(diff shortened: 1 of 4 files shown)


PDegenPortnoy:
This looks very good and the tests seem comprehensive. Good to merge & deploy

PR #1433, opened 2019-11-19
OTWO-5751: update 'rubocop' gem with 'rubocop-rails'

bostonpops:
This PR addresses replacing the existing rubocop gem with rubocop-rails, and correcting any issues found by the new gem.  Refer to https://jira-sig.internal.synopsys.com/browse/OTWO-5751.

Rubocop has moved Rails cops to separate gem 'rubocop-rails'. So, we need to replace 'rubocop' gem with 'rubocop-rails'.

`rubocop-rails` identified the following offending cops:

- `Rails/UnknownEnv`:
resolved

- `Rails/HasAndBelongsToMany`:
resolved

- `Rails/EnumHash`:
resolved

- `Rails/SkipsModelValidations`:
Rubocop prefers using update_attributes as opposed to update_attribute or update_column as the former will perform validations.  This cop is currently `disabled` in `.rubocop.yml`.

- `Rails/DynamicFindBy`:
Rubocop discourages use of the dynamic find_by_* methods, and suggest using find_by instead.  This cop is currently `disabled` in `.rubocop.yml`.

- `Rails/OutputSafety`:
Rails discourages use of html_safe and raw.
In our code, there are many strings with embedded html that require calling html_safe. To address this I've created a single method, safe_text, that accepts a string of chars, calls html_safe and returns the resulting string. safe_text is now the only code that calls html_safe, and the OutputSafety cop is disabled for this method.

- `Rails/InverseOf`:
Rubocop throws this violation when Active Record can't automatically determine the inverse association because of a scope or the options used.  This cop is currently `disabled` in `.rubocop.yml`.

- `Rails/HasManyOrHasOneDependent`:
Rubocop wants has_many and has_one associations to specify a :dependent option. It doesn't register an offense if :through option was specified.  This cop is currently `disabled` in `.rubocop.yml`.

- `Rails/HelperInstanceVariable`:
Rubocop discourages the use of instance variables in helper classes.  This cop is currently `disabled` in `.rubocop.yml`.

- `Rails/LexicallyScopedActionFilter`:
Rails wants all methods referenced a filter's only or except options to be present in the implementing class.  I resolved this copy for simple ActionController methods (ie: create,new,index,show,edit,update,delete,destroy); and turned the cop off, where necessary, for any other non-RESTful methods.

### DIFF
--- a/app/controllers/concerns/enlistment_filters.rb
+++ b/app/controllers/concerns/enlistment_filters.rb
@@ -17,20 +17,6 @@ module EnlistmentFilters
     before_action :project_has_code_location?, only: :create
   end
 
-  def create; end
-
-  def new; end
-
-  def index; end
-
-  def show; end
-
-  def edit; end
-
-  def update; end
-
-  def destroy; end
-
   private
 
   def enlistment_params
--- a/app/controllers/concerns/enlistment_filters.rb
+++ b/app/controllers/concerns/enlistment_filters.rb
@@ -17,6 +17,20 @@ module EnlistmentFilters
     before_action :project_has_code_location?, only: :create
   end
 
+  def create; end
+
+  def new; end
+
+  def index; end
+
+  def show; end
+
+  def edit; end
+
+  def update; end
+
+  def destroy; end
+
   private
 
   def enlistment_params
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -13,7 +13,7 @@ module ApplicationHelper
     err = model.errors[attr]
     return '' if err.blank?
 
-    haml_tag 'p', safe_text([err].flatten.join('<br />')), opts.reverse_merge(class: 'error').merge(rel: attr)
+    haml_tag 'p', [err].flatten.join('<br />').html_safe, opts.reverse_merge(class: 'error').merge(rel: attr)
   end
 
   def password_error_tag(model, attr, opts = {})
@@ -21,7 +21,7 @@ module ApplicationHelper
     return '' if err.blank? || err == ["can't be blank"]
 
     err = err.first if err.size == 2
-    haml_tag 'p', safe_text([err].flatten.join('<br />')), opts.reverse_merge(class: 'error').merge(rel: attr)
+    haml_tag 'p', [err].flatten.join('<br />').html_safe, opts.reverse_merge(class: 'error').merge(rel: attr)
   end
 
   def project_pages_title(page_name = nil, project_name = nil)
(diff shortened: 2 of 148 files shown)


app/helpers/application_helper.rb
```diff
@@ -138,6 +138,12 @@ def api_pagination(response)
     will_paginate(numbers.paginate(page: response['current_page'], per_page: response['per_page']))
   end
 
+  # rubocop: disable Rails/OutputSafety
+  def safe_text(str)
+    safe_join(Array(str.html_safe))
+  end
+  # rubocop: enable Rails/OutputSafety
```
PDegenPortnoy: I have grave concerns about this. `html_safe` was replaced with this method and the two are doing polar opposite things.

[html_safe](https://apidock.com/rails/String/html_safe) marks the reference string as *being* safe and returns an instance of a `SafeBuffer`. "It will be inserted into HTML with no additional escaping performed"

[safe_join](https://apidock.com/rails/ActionView/Helpers/OutputSafetyHelper/safe_join), while also returning an HTML safe string (not sure if it is an instance of a `SafeBuffer` or not), also "all items ... are HTML escaped"

The first method *does not* HTML escape, but the second one does. These two things are not equivalent.

I see the [Rubocop documentation](https://www.rubydoc.info/gems/rubocop-rails/2.3.2/RuboCop/Cop/Rails/OutputSafety), but don't know if it's a simple as a blanket replacement.

app/controllers/concerns/enlistment_filters.rb
```diff
@@ -17,6 +17,20 @@ module EnlistmentFilters
...
+  def index; end
+
+  def show; end
+
+  def edit; end
+
+  def update; end
+
+  def destroy; end
+
```
PDegenPortnoy: Why does Rubocop require empty methods in a Concern? This is weird and just looks wrong to have this empty, dead code.

PDegenPortnoy:
Also, we really need all these separate commits squashed. :-)

app/helpers/application_helper.rb
bostonpops: Appreciate the comments on `safe_text` and `html_safe`/`safe_join`.  I was trying to create a single method that would be responsible for calling `html_safe` and then, using `safe_join` any unescaped HTML would then be escaped, addressing any potential malicious string text.
Agreed that it's not as simple as a blanket replacement, we'll back this change out and turn the cop off (`Rails/OutputSafety`) in the `.rubocop.yml` file.

I also agree with removing the dead code (stubbing restful controller methods) from the respective controllers.  Hindsight tells me this code shouldn't have been as it doesn't add much.  I think Rubocop likes the idea of having every controller method defined so long as it is being referenced somewhere else within the controller, it's definitely a convenience but  all that dead code distracts from grokking the real functionality; let's not try and outthink Rubocop here, we'll turn the cop off (`Rails/LexicallyScopedActionFilter`) in `.rubocop.yml`.

bostonpops: Re testing:  My test environment runs cleanly, not sure why you'd be seeing the VCR errors (and others) but duly noted, and hope this doesn't start happening consistently.

PR #1434, opened 2019-11-20
OTWO-5802 Fixed undefined method for nil commit_contributor

Priya5:


PDegenPortnoy:
A reasonable short-term fix; please merge and push into production

PR #1435, opened 2019-11-20
OTWO-5755: OH Home Page, Update CSS

bostonpops:
This PR addresses the updates to the OH Home Page CSS regarding the new OTWO color scheme.  Refer to JIRA ticket: https://jira-sig.internal.synopsys.com/browse/OTWO-5755.

Refer to `app/assets/stylesheets/oh-colors.sass` for the implementation of the new OTWO color scheme.  It uses a Sass `Map` (`$COLORS`) to manage new colors both with and without gradient definitions.  Gradient definitions also utilize a Sass `Map` and currently provide values for `120%`, `100%`, `80%`, `60%`, `%40%`, and `20%`.  Since Sass does not allow multiline variable definitions, the definition of the `$COLORS` map must be kept on one line, hence the naming of the keys and value used to minify the definition. 

The Sass `@function`, `color-gradient`, takes a `$color` argument to be used as the main key to the `$COLORS` map.  A $gradient argument is also specified, should the color be represented with the necessary gradient definitions.  In the event that a color does not have a gradient definition, a second argument of '0' (actually, any integer value) needs to be provided as Sass does not allow for variable argument lists for functions.  Examples:
```
h2
    background-color: color-gradient($SECTION_BACKGROUND, 0)   // no gradient val necessary

a
    color: color-gradient($PRIMARY_PURPLE, 100)   // render '#5A2A82' for 100% Primary Purple
```


PDegenPortnoy:
Shouldn't this be merging into the OTWO-5664 branch, and not master? OTWO-5664 is the Epic and should be the feature branch. We're not ready to have the UI updates merge into master. I apologize if I muddied the ideas this morning during standup.

bostonpops:
Ah, you're right.  I overlooked that when creating the PR, but knew that these commits were to be squashed and merged into the feature branch (OTWO-5664, as you've said above).

I have a few updates to add per our conversation this AM, upon which the plan still is to squash commits and merge to OTWO-5664 (feature branch).  All we need now is the internet.  :/

PR #1441, opened 2019-12-05
OTWO-5831: Fix invalid version patterns breaking project#security

bostonpops:
This PR fixes errors that block a user from accessing projects#security.

The error occurs when sorting the version names that comprise the dropdown in the Vulnerabilities section of the Projects Security page, and is related to non-standard semantic version patterns used within a number of projects hosted on OpenHub. 

The first fix recognizes non-standard chars in version names, such that `nil` is no longer assigned for non-standard chars in the array to be sorted, which was causing `comparison of Array with Array failed`

The second fix checks for a `nil` token passed to `get_token_array` and simply returns an empty array (`[]`).  Release names that ended with a hyphen (ie: `TestRelease-.x.y.z`) were found to cause this issue.

Tests have been augmented such that release names which were causing previous errors, now pass with every build.

### DIFF
--- a/app/helpers/vulnerabilities_helper.rb
+++ b/app/helpers/vulnerabilities_helper.rb
@@ -145,6 +145,8 @@ module VulnerabilitiesHelper
   end
 
   def get_token_array(token)
+    return [] if token.nil?
+
     if token.include?('-')
       token1, token2 = token.split('-')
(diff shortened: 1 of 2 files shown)


PDegenPortnoy:
Looks good to me. +1 to merge

PR #1460, opened 2020-01-06
OTWO-5855: fix encoding issue on links index page

karthikeyanBD:


app/helpers/links_helper.rb
```diff
@@ -1,9 +1,12 @@
...
+
 module LinksHelper
   def safe_slice_host(url, length = 33)
-    safe_url = sanitize(url)
+    safe_url = sanitize(URI.encode(url))
     hostname = URI.parse(safe_url).host
     truncate(hostname, length: length)
   end
 end
+# rubocop: enable Lint/UriEscapeUnescape
```
notalex: rubocop considers URI.encode as bad practice. We need some research into why it is so. Do not merge until we get some information on this.

karthikeyanBD: Sure take your time.

karthikeyanBD:
@notalex We can't use `sting#encode` method because it's not encoding non-ASCII characters, it just removes it.

```ruby
url = "http://en.wikipedia.org/wiki/Étoilé"
url.encode('binary', 'UTF-8', invalid: :replace, undef: :replace)
"http://en.wikipedia.org/wiki/?toil?"
```

On the other hand `URI#encode` is capable of encoding non-ASCII characters. That's what we are looking for.
```ruby
URI.encode(url)
"http://en.wikipedia.org/wiki/%C3%89toil%C3%A9"
```

notalex:
@karthikeyanBD, we just want to extract the host url here. So we do not want to worry about characters following the base url. The host url will always be ascii, so *URI.encode* will never obfuscate that part.

app/helpers/links_helper.rb
notalex: URI.encode/escape is deprecated due to certain [bugs](https://bugs.ruby-lang.org/issues/9806). Also see [this](https://stackoverflow.com/questions/2824126/whats-the-difference-between-uri-escape-and-cgi-escape) discussion.

PDegenPortnoy:
This has been open a while; what do we want to do with it?

PR #1466, opened 2020-01-30
OTWO-5834 Fixed incompatible encoding issue

Priya5:


PDegenPortnoy:
Seems straightforward and clean. Should we have a test case to cover this?

Priya5:
The record which has caused the error is invalid. We are not able to regenerate the tag with invalid encoding because of validation on tag#name.
> [8] pry(main)> t = Tag.find 129380
  Tag Load (294.7ms)  SELECT  "tags".* FROM "tags" WHERE "tags"."id" = $1 LIMIT 1  [["id", 129380]]
=> #<Tag:0x00007fa4a3592028 id: 129380, name: "simple´", taggings_count: 1, weight: 1.0>
[9] pry(main)> t.valid?
=> false
[10] pry(main)> t.errors
=> #<ActiveModel::Errors:0x00007fa4868312b0 @base=#<Tag:0x00007fa4a3592028 id: 129380, name: "simple´", taggings_count: 1, weight: 1.0>, @messages={:name=>["may include only letters, numbers, +, and _"]}>

PR #1479, opened 2020-02-19
OTWO-5765: Update CSS for the Project Commits pages

bostonpops:
This PR addresses updating the colors defined for the `commits#index` and `commits#summary` pages, with respect to the new UI Trade Dress. 

A few issues are still open before squashing and merging to the feature branch:
1) color updates for the range buttons on the Commits Per Month chart
2) color definition for the Commits Per Month chart
3) should the "Month Recent Commits" heading be revised to "Most Recent Commits"?

### DIFF
--- a/app/assets/stylesheets/fontawesome_icons.sass
+++ b/app/assets/stylesheets/fontawesome_icons.sass
@@ -1,5 +1,5 @@
 [class^="icon-"], [class*=" icon-"]
-  font-family: FontAwesome
+  font-family: FontAwesome !important
   font-weight: normal
   font-style: normal
   text-decoration: inherit
--- a/app/assets/stylesheets/base.sass
+++ b/app/assets/stylesheets/base.sass
@@ -16,12 +16,17 @@
 =link_blue
   @include site-link-color
 
+// font awesome custom styles
+#commits_index_page,
+#commits_summary_page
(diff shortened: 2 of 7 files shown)


app/assets/stylesheets/fontawesome_icons.sass
```diff
@@ -1,5 +1,4 @@
 [class^="icon-"], [class*=" icon-"]
-  font-family: FontAwesome
```
rapbhan: are we sure about this? wouldn't it affect across all the pages if it's removed?

bostonpops: it would, good point.   We are trying to use Roboto font across all pages, but, agreed that this may introduce unintended behavior elsewhere.  will refactor so that the styling for the commits pages is separate from the global styling.

app/assets/stylesheets/fontawesome_icons.sass
```diff
@@ -24,6 +24,23 @@
   vertical-align: -10%
   font-size: 1.3333333333333333em
 
+#commits_index_page,
```
rapbhan: Could we please move the custom styles to a different file (new or existing one) such that in the future when we upgrade the font-awesome lib, the changes wouldn't be lost.

PR #1492, opened 2020-04-13
OTWO-5945 Updated sql scripts

Priya5:


PDegenPortnoy:
Just to confirm; we need these scripts to drop functions that are no longer needed because we are no longer using the FDW, correct? And this change is to ensure we access the functions in the correct schema after the `pg_restore`

Should we consider dropping these functions before we do the `pg_dump`?

Priya5:
Yes, We do not need those functions, views and sequences after removing FDW. 
We should drop these functions after `pg_restore` so that we will have backup of original database.

PR #1490, opened 2020-03-23
OTWO-5945 Removed secondbase gem and updated structure.sql

Priya5:

PR #1494, opened 2020-04-20
Revert "OTWO-5945 Removed secondbase gem and updated structure.sql"

PDegenPortnoy:
Reverts blackducksoftware/ohloh-ui#1490. We need to keep this feature branch separate until we are ready to move the OH DB to the FIS DB server


msk999:
No longer needed

PR #1513, opened 2020-09-02
OTWO-6101 Sanitize page title to prevent JS injections

notalex:


PDegenPortnoy:
Are there any other layout base-pages that might have this issue as well? We’d like to ensure a systemic fix is possible.

notalex:
Updating the JIRA ticket for more details.

PR #1519, opened 2020-10-15
OTWO-6221 Fixed implicit conversion of String into Integer

Priya5:


app/controllers/concerns/enlistment_filters.rb
```diff
@@ -101,7 +101,7 @@ def enlistment_deleted?
 
   def get_code_location_id
     response = CodeLocationApi.new(url: @code_location.url, branch: @code_location.branch).fetch
-    JSON.parse(response).first['id'] unless response.nil?
+    JSON.parse(response)['id'] unless response.nil?
```
PDegenPortnoy, line 104: This looks pretty important; like the eFISbot Code Location API changed its return type from an array to a single instance. Is that correct? If so, when did we change the Code Location API and is it having an impact on the KB Discovery code? If there is a chance that this end point will return an array, then we should convert all responses to an array so they are treated the same: `Array(response)` has no effect if `response` is already an array

Priya5: The eFisbot Code Location API returns single instance only when url and branch is present. https://github.com/blackducksoftware/fisbot/blob/master/app/controllers/code_locations_controller.rb#L9

PDegenPortnoy: OK, that makes sense. But it would seem that the API changed because the code used to handle receiving an array of responses. I agree that the Code Location `new` end point (with a URL and branch) should return a single Code Location, but am unclear as to why it was ever anything else in this method. I understand that there are other calls to the API that might return an array, but in this method, we should have always expected a single result. Carry on :-) Good to merge.

PR #1533, opened 2021-02-09
Added Sql tracker to track sloc_metrics table

karthikeyanBD:


config/environments/production.rb
```diff
@@ -84,4 +84,8 @@
 
   # Do not dump schema after migrations.
   config.active_record.dump_schema_after_migration = false
+
+  SqlTracker::Config.enabled = true
```
sig-krasnick, line 88: Let's change this to the environment variable we used for the other apps

config/environments/staging.rb
```diff
@@ -86,4 +86,8 @@
 
   # Do not dump schema after migrations.
   config.active_record.dump_schema_after_migration = false
+
+  SqlTracker::Config.enabled = true
```
sig-krasnick, line 90: Here too

sig-krasnick (changes requested):
Environment variable changes

PR #1536, opened 2021-03-01
OTWO-6359 Changed application smtp configurations

Priya5:


config/environments/staging.rb
```diff
@@ -68,7 +68,7 @@
   config.action_mailer.raise_delivery_errors = false
   # config.action_mailer.perform_deliveries = true
   config.action_mailer.delivery_method = :letter_opener_web
-  config.action_mailer.smtp_settings = { address: 'mailrelay.blackducksoftware.com',
+  config.action_mailer.smtp_settings = { address: 'mailhost.synopsys.com',
```
sig-krasnick, line 71: This should remain as mailrelay.blackducksoftware.com for staging.  Not sure this will be valid, but it shouldn't go through the production mail server.

bdsdeployer (commented):
I have been informed that the correct mail server is dmz-smtprelay.synopsys.com.  This needs to be put in today, since the old email server is being blocked.

PR #1547, opened 2021-06-10
OTWO-6352 - storing smtp address into ENV variable

sig-krasnick:


sig-krasnick:
Storing SMTP address in env variable and rolling back last commit for now.

PR #1550, opened 2021-07-07
OTWO-6433 Script to add fedora code location

Priya5:


sig-krasnick:
This code will be replaced with API work being done.

PR #1553, opened 2021-07-27
OTWO-6437 - ohloh-ui changes for GCP migration

sig-krasnick:


config/initializers/airbrake.rb
```diff
@@ -3,7 +3,8 @@
 Airbrake.configure do |config|
   config.environment = Rails.env # must be set for ignore_environments to work
   config.project_key = ENV['AIRBRAKE_API_KEY']
-  config.host = "#{ENV['AIRBRAKE_HOST']}:#{ENV['AIRBRAKE_PORT'].to_i}"
+  config.host = "#{ENV['AIRBRAKE_HOST']}:#{ENV['AIRBRAKE_PORT'].to_i}" unless ENV['KUBERNETES_PORT']
+  config.host = ENV['AIRBRAKE_HOST'] if ENV['KUBERNETES_PORT']
```
notalex, line 7: We can instead do:

```ruby
config.host = "#{ENV['AIRBRAKE_HOST']}#{':' + ENV['AIRBRAKE_PORT'] unless ENV['KUBERNETES_PORT']}"
```

Try it out in irb. Won't work if port is a non string value, which is unlikely since ENV value will be a string.

```ruby
env = { 'host' => 'host', 'port' => 'port' }
p "#{env['host']}#{':' + env['port'] unless false}"
```

sig-krasnick: yes, that is cleaner

PR #1555, opened 2021-08-03
OTWO-6417 Allow whitelisting of folders

Priya5:


app/models/analysis.rb
```diff
@@ -75,6 +75,12 @@ def ignore_tuples
     end.compact.join(' AND ')
   end
 
+  def allowed_tuples
+    [].tap do |tuples|
+      analysis_sloc_sets.each { |analysis_sloc_set| tuples << analysis_sloc_set.allowed_tuples }
```
notalex, line 80: This can be simplified as discussed before.

app/views/enlistments/_enlistment.html.haml
```diff
@@ -19,10 +19,22 @@
...
+        = t('.description4')
+        = pluralize(ass.allowed_fyle_count, t('.file'))
+        = t('.description2')
   %td.col-md-2.last
     - if has_permission
       = icon_button(edit_project_enlistment_path(@project, enlistment), icon: 'pencil',
         text: t('.edit_ignored'), size: 'mini', type: 'primary')
+      - if current_user_is_admin?
+        = icon_button(edit_allowed_files_project_enlistment_path(@project, enlistment), icon: 'pencil',
+          text: t('.edit_allowed'), size: 'mini', type: 'primary')
```
notalex, line 37: This should be a feature open to managers. So place this under **has_permission** block above it.

lib/allowed.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+module Allowed
+  def self.parse(text)
+    text.to_s.each_line.map do |line|
+      line.to_s.strip =~ /^\s*([^#\*\s][^\*\s]+)\*?(\s+|$)/i
+      Regexp.last_match(1)
+    end.compact
```
notalex, line 8: This can be simplified with a `split("\n")`

notalex (changes requested):

notalex:
LGTM

PR #1557, opened 2021-08-23
OTWO-6476 - Unable to access new project's show page

sig-krasnick:


lib/openhub_security.rb
```diff
@@ -1,26 +1,47 @@
...
+    http = get_http(url)
+
+    request = get_http_request(url)
+    begin
+      response = http.request(request)
+      response = JSON.parse(response.read_body)
+    rescue SocketError => e
+      Rails.logger.error("OpenhubSecurity SocketError Exception:  #{e.message}")
+    end
+    response
```
notalex, line 45: Does response need a default value? It seems to gets overriden on assignment in every scenario.
  The above function can be simplied.
  ```
  def self.get_response(url)
    http = get_http(url)

    request = get_http_request(url)
    response = http.request(request)
    response = JSON.parse(response.read_body)
  rescue SocketError => e
    Rails.logger.error("OpenhubSecurity SocketError Exception:  #{e.message}")
    response
  end
  ```

notalex (changes requested):

lib/openhub_security.rb
notalex: If the intention was to catch only the SocketError during `http.request`, then let's extract it into another method.

```ruby
  def self.get_response(url)
    http = get_http(url)

    request = get_http_request(url)
    get_safe_response(http, request)
  end

  def self.get_safe_response(http, request)
    response = http.request(request)
    response = JSON.parse(response.read_body)
  rescue SocketError => e
    Rails.logger.error("OpenhubSecurity SocketError Exception:  #{e.message}")
    response || {}
  end
```

notalex:
LGTM. The logic is accurate and the tests are correct.

test/controllers/projects_controller_test.rb
```diff
@@ -464,6 +464,14 @@
       must_respond_with :ok
     end
 
+    it 'should throw SocketError when getting UUID from BlackDuck KB and proceed to show page' do
+      Net::HTTP.any_instance.expects(:request).raises(SocketError)
+      project = create(:project, uuid: '', name: 'rails')
+      get :show, id: project.to_param
+      _(project.reload.uuid).must_be_empty
+      must_respond_with :ok
+    end
```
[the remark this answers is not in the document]
sig-krasnick: I kept the SocketError specific description for clarity of what exception is being tested.

lib/openhub_security.rb
sig-krasnick: Thanks!

PR #1569, opened 2021-11-02
OTWO-6500 Replace NewRelic with Datadog

notalex:


test/controllers/application_controller_test.rb
```diff
@@ -108,7 +108,7 @@
 
     it 'reports to newrelic on FisbotApiError' do
```
sig-krasnick, line 109: Change newrelic comment to datadog

notalex: Good find. I grepped for newrelic and found even more places to edit.

PR #1584, opened 2022-02-16
Fix api enlistment controllers to accept nil branch

BlackDuckDave:


sig-krasnick:
retest this please

BlackDuckDave:
The tests all pass, but there is an error in the Simplecov gem that is causing the checks to fail

PR #1587, opened 2022-02-28
OTWO-6570 Handle errors for missing template formats

notalex:


notalex:
The build fails due to an error within *simpleCov* gem. Everything above it passes which includes rubocop and tests. We may merge this PR.

Priya5 (approved):

PR #1588, opened 2022-03-02
OTWO-6620 Added scan_project_id to projects

Priya5:


notalex:
Please rename `scan` to `coverity` in DB column and variable names. The name **scan** is too generic and might not make immediate sense when we look at this in the future.

PR #1591, opened 2022-03-10
Cleanup unused test & development gems

notalex:
binding_of_caller -> Binding.of_caller is not used anywhere.
guard -> If one needs guard, they can install it individually.
meta_request -> Used for a chrome extension
flog -> Install this locally if needed. We do not need it in Gemfile.
m -> Rails 5 already provides a way to run tests by line number.
ruby_parser -> we do not use this anywhere.
spring -> raises exceptions related to autoload on running tests.
capybara, selenium, jasmine & spinach -> No integration tests right now.
pry -> Install this locally if needed. We do not need it in Gemfile.
rails-erd -> Install this locally when needed.
rb-readline -> This is not used anywhere.
rubocop-faker -> We do not call rubocop with this extension.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -55,7 +55,6 @@ gem 'will_paginate-bootstrap'
 
 group :development do
   gem 'better_errors'
-  gem 'binding_of_caller'
   gem 'bootsnap', '~> 1.10.3', require: false
   gem 'capistrano'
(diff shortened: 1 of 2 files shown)


Priya5 (approved):

notalex:
retest this please

PR #1592, opened 2022-03-13
Revert "OTWO-6620 Added scan_project_id to projects"

PinkDiamond1:
Reverts blackducksoftware/ohloh-ui#1588


bdsdeployer:
Can one of the admins verify this patch?

notalex:
@PinkDiamond1, if this is breaking your use case, please let us know. We can help.

PR #1594, opened 2022-03-16
Remove unused gems from production

notalex:
execjs is a dependency of uglifier, so we needn't specify it.
ohloh_scm is not used anymore. It is a vestige from old ohloh source.
therubyracer js runtime is replaced by nodejs installed on server.


notalex:
retest this please

PR #1600, opened 2022-04-06
OTWO link in admin page to accounts page

Niharika1117:


app/assets/javascripts/charts.js
```diff
@@ -52,9 +52,17 @@ var Charts = {
     }
 
     if (data.xAxis) {
-      $.extend(data.xAxis.labels, {formatter: function(){
-        return this.value.stringify;
-      }});
+        $.extend(data.xAxis.labels, {formatter: function(){
+          var first_day = new Date(this.value)
+          var last_day = new Date(first_day.getFullYear(), first_day.getMonth()+1, 0);
+          var month = ['Jan', 'Feb', 'Mar', 'Apr', 'May', 'Jun', 'Jul', 'Aug', 'Sep', 'Oct', 'Nov', 'Dec']
```
sig-krasnick, line 58: This should be months instead of month for clarity.

notalex:
The commits message is missing the ticket number.

app/assets/javascripts/charts.js
sig-krasnick: Please use the full Jira ticket number, i.e. OTWO-xxxx in your commits so we have an association between this work and the ticket.

sig-krasnick:
Be careful not to use git force pushed.  The --force option for git push allows you to override this rule: the commit history on the remote will be forcefully overwritten with your own local history. This is a rather dangerous process, because it's very easy to overwrite (and thereby lose) commits from your colleagues.

PR #1601, opened 2022-04-12
OTWO-6666 Project Trend charts

Niharika1117:


app/assets/javascripts/project.js
```diff
@@ -0,0 +1,64 @@
+var ProjectDashboard = {
+  init: function() {
+    $("#one_year_project, #six_months_project, #three_months_monthly_project, #six_months_monthly_project, #one_year_monthly_project").hide();
+    $('input[name="options_project"]').change( function() {
+      console.log("hiiiiii");
```
sig-krasnick, line 5: remove console message

app/assets/javascripts/project.js
```diff
@@ -0,0 +1,64 @@
...
+      ProjectDashboard.update_chart($(this).val());
+    })
+    $('input[name="radio_project"]').change( function() {
+      ProjectDashboard.update_chart($(this).val());
+    })
+  },
+  update_chart: function(months) {
+    if(months == '3' && $('input[name="radio_project"]').filter(':checked').val() == 'weekly') {
```
sig-krasnick, line 13: This function is too long and repetitive in general.  It should be broken into at least 2 methods.  First to determine the selection and the second to handle the showing and hiding of components.  Pass whatever you need for the appropriate show into this new method

app/decorators/oh_admin/project_chart.rb
```diff
@@ -0,0 +1,111 @@
...
+
+  private
+
+  def set_series_data(chart)
+    chart['series'][0][:data] = @analyzed.values
+    chart['series'][1][:data] = @non_analyzed.values
+    chart['series'][2][:data] = @total_count.flatten
+  end
+
+  def account_data
```
sig-krasnick, line 29: This should be changed to project_data

app/decorators/oh_admin/project_chart.rb
```diff
@@ -0,0 +1,111 @@
...
+    else
+      Project.left_joins(:best_analysis).group('date(projects.created_at)')
+             .where('projects.created_at >= ? AND projects.created_at <= ? AND projects.best_analysis_id is NULL',
+                    @from, @to).references(:best_analysis).count
+    end
+  end
+
+  def total_projects(date)
+    if @filter == 'monthly'
+      Project.left_joins(:best_analysis).group("DATE_TRUNC('month', projects.created_at)")
```
sig-krasnick, line 102: We shouldn't need to left join with analysis here to get the total projects

sig-krasnick (changes requested):

Niharika1117:
Retest this please

app/assets/javascripts/project.js
```diff
@@ -1,38 +1,24 @@
...
-    }
-    else if(months == '6' && filter == 'monthly') {
-      $("#six_months_monthly_project").show();
-      $("#three_months_project, #six_months_project, #one_year_project, #three_months_monthly_project, #one_year_monthly_project").hide();
-    }
-    else if(months == '12' && filter == 'monthly') {
-      $("#one_year_monthly_project").show();
-      $("#three_months_project, #six_months_project, #one_year_project, #three_months_monthly_project, #six_months_monthly_project").hide();
+    else if(filter == 'weekly') {
+      var contClass = $('input[name="project_options"]').filter(':checked').data('div');
```
sig-krasnick, line 20: This line can be called once and passed into the function.  Otherwise looks good

Niharika1117: I tried calling that the line at once, But problem is button checked value is changing dynamically and when we click on button change function is getting called.

So, both the change function I need to call that line again.

sig-krasnick (approved):

PR #1605, opened 2022-04-26
OTWO-6677 Fix Accounts#commits_by_project chart

notalex:


Priya5 (approved):

sig-krasnick (commented):
Any reason not to remove the entire function?  It doesn't do anything.

PR #1606, opened 2022-04-28
OTWO-6613 New code location redirection issue

Niharika1117:


app/controllers/enlistments_controller.rb
```diff
@@ -32,7 +33,7 @@ def create
       redirect_to project_enlistments_path(@project)
     else
       flash[:error] = @code_location.errors['error']
-      render :new, status: :unprocessable_entity
```
sig-krasnick, line 35: Why does the status need to be changed?

test/helpers/projects_helper_test.rb
```diff
@@ -146,7 +146,7 @@ class ProjectsHelperTest < ActionView::TestCase
     it 'should return project managers list' do
       project_manager = create(:manage)
       @project = project_manager.target
-      _(project_managers_list).must_match project_manager.account.name
+      _(project_managers_list).to_h[:target].match(/(a href|\/\/).*/)
```
sig-krasnick, line 149: Can I get an explanation for the change from a specific account name to a regex on an href?

app/controllers/enlistments_controller.rb
Niharika1117: I just checked other places like review_controller, We dont use this status as we render new link on unauthorization.

So, just to maintain the code consistency , I have also removed the status code.

test/helpers/projects_helper_test.rb
Niharika1117: Previously we were using the disabled_button code from buttons_helper , which was not using the link for redirection.

So, this test case started failing because instead of name , Now its coming link in the account name.

So , I modified the test case according to that.

sig-krasnick (approved):

PR #1608, opened 2022-04-28
OTWO-6656 Write exceptions to file

notalex:
Datadog & Errbit are not working right now

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -277,7 +277,11 @@ class ApplicationController < ActionController::Base
     if ENV['KUBERNETES_PORT']
       notify_airbrake(exception)
     else
-      DataDogReport.error(exception.backtrace.to_s)
+      logger = Logger.new(Rails.root.join('log/errors.log'))
+      logger.error(exception.message)
+      logger.error(exception.backtrace)
+      logger.error('-' * 150)
+      logger.close
     end
   end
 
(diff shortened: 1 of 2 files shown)


app/controllers/application_controller.rb
```diff
@@ -277,7 +277,11 @@ def report_errors(exception)
     if ENV['KUBERNETES_PORT']
       notify_airbrake(exception)
     else
-      DataDogReport.error(exception.backtrace.to_s)
+      File.open('log/errors.log', 'a') do |file|
```
sig-krasnick, line 280: We should be using Rails.logger.error instead.  That mechanism should already be in place.

sig-krasnick: We can configure the format of the logs as shown here https://stackify.com/rails-logger-and-rails-logging-best-practices/

notalex: Simply using **Rails.logger** will log to the **production.log**. We wish to use a separate file to keeps things manageable.
To use the Logger class and write to a separate file, we have to create a new **Logger** object.
```
@logger = Logger.new('...new_file')
@logger.info(...)
```
We are avoiding this here because it doesn't provide any additional value. The primary use for Logger is to control log output using **info**, **debug** etc.
In current case, we are not limiting the log. This function is only called in non development environment and we wish to log everything then.

PR #1610, opened 2022-05-17
OTWO-6694 Add an endpoint for health check

notalex:


notalex:
Added feature in another email.

PR #1616, opened 2022-05-25
OTWO-6700 Admin page improvement

Niharika1117:


app/assets/javascripts/charts.js
```diff
@@ -56,13 +56,12 @@ var Charts = {
...
-                    // this.value + '</a>';
-        // else
-          // return '<a href="/admin/accounts?commit=Filter&q[created_at_gteq_datetime]=' + this.value + '&q[created_at_lteq_datetime]=' + this.value + '" target="_blank">' +
-                    // this.value + '</a>';
+        if (months.includes(this.value.split(' ')[0]))
+          return '<a href="/admin/accounts?commit=Filter&q[created_at_gteq_datetime]=' + first_day + '&q[created_at_lteq_datetime]=' + last_day + '" target="_blank">' +
+                    this.value + '</a>';
+        else
+          return '<a href="/admin/accounts?commit=Filter&q[created_at_gteq_datetime]=' + this.value + '&q[created_at_lteq_datetime]=' + this.value + '" target="_blank">' +
+                    this.value + '</a>';
```
Niharika1117, line 64: The above code was commented out on the Main branch. This code has been done on OTWO-6652 providing the link to the account trend graph which will go to the account page.

So added this one with this commit only.

app/models/concerns/project_scopes.rb
```diff
@@ -48,6 +48,6 @@ module ProjectScopes
                  .having('count(*) >= ?', tags.split.flatten.length)
     }
     scope :with_analysis, -> { active.where.not(best_analysis_id: nil) }
-    scope :active_enlistments, -> { active.joins(:enlistments).where(enlistments: { deleted: false }) }
+    scope :active_enlistments, -> { active.joins(:enlistments) }
```
Niharika1117, line 51: Removed the where clause because enlistment is having default scope with the same one.

So adding the where clause was adding two times the same conditions.

app/helpers/dashboard_helper.rb
```diff
@@ -47,10 +47,6 @@ def outdated_projects
     number_to_percentage((projects_count.to_f / active_projects_count) * 100, precision: 2)
   end
 
-  def convert_to_datetime(value)
-    Time.current.ago(value).utc if value
-  end
-
```
Niharika1117, line 50: This Method was not in use.

Priya5:
Merging this PR to fix GCP Jenkins build.

PR #1620, opened 2022-06-01
OTWO-6706 removed KUBERENETS_PORT from code

Niharika1117:


test/lib/data_dog_report_test.rb
```diff
@@ -6,7 +6,7 @@ class DataDogReportTest < ActiveSupport::TestCase
   it 'must send a create_event request to Datadog' do
     VCR.use_cassette('datadog_error_request') do
       response = DataDogReport.error('This is an error sample')
-      _(response.to_hash[:status]).must_equal 'ok'
+      _(response).must_equal nil
```
sig-krasnick, line 9: This is a strange test.  Why did this change from 200 status to no status?

sig-krasnick (commented):
This looks fine.  One comment regarding the status code of Datadog error request should be looked at.

test/controllers/application_controller_test.rb
```diff
@@ -113,8 +108,7 @@ class ApplicationControllerTest < ActionController::TestCase
 
     it 'must log errors to file when not on kubernetes' do
       Rails.application.config.stubs(:consider_all_requests_local).returns false
-      ENV.delete('KUBERNETES_PORT')
-      Logger.any_instance.expects(:error).times(3)
+      Logger.any_instance.expects(:error).never
       get :throws_standard_error
       assert_response :not_found
       Rails.application.config.unstub(:consider_all_requests_local)
```
notalex, line 114: Remove this test. It is no longer applicable.

test/lib/data_dog_report_test.rb
notalex: Find a way to the **puts** message to stdout instead. No need to check that response is nil.

notalex (changes requested):

sig-krasnick (approved):

PR #1639, opened 2022-07-20
OTWO-6739 create Spam review workflow

BlackDuckDave:
Initial commit, create migration, simple API endpoint, test


app/controllers/spam_controller.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+class SpamController < ApplicationController
+  before_action :session_required, only: %i[redirect_to_first_potential_spammer]
+  before_action :admin_session_required, only: %i[redirect_to_first_potential_spammer]
+
+  def redirect_to_first_potential_spammer
+    sql = <<-SQL.squish
+             SELECT id FROM oh.potential_spammers LIMIT 1;
+    SQL
```
notalex, line 10: This can be simplified to a single line:
```
sql = 'Select id FROM...'
```

db/migrate/20220718210823_create_reviewed_not_spammer.rb
```diff
@@ -0,0 +1,26 @@
...
+
+class CreateReviewedNotSpammer < ActiveRecord::Migration[5.2]
+  def up
+    create_table :reviewed_not_spammers do |t|
+      t.references :account, foreign_key: true, unique: true
+      t.timestamps
+    end
+
+    execute <<-SQL.squish
+      CREATE VIEW potential_spammers AS
```
notalex, line 11: Will this work without defining it as **oh.potential_spammers** ?

config/locales/accounts.en.yml
```diff
@@ -143,6 +143,10 @@ en:
           link_html: '<i class="icon-check">&nbsp;</i>MARK AS SPAMMER LEVEL'
           confirm: 'Are you sure you want to mark %{name} as a spammer, disabling the account and removing all posts?'
           title: 'Mark %{name} as spammer!'
+        not_spammer:
+          link_html: '<i class="icon-check">&nbsp;</i>CONFIRM NOT SPAMMER'
+          confirm: 'Are you sure you want to mark %{name} as not a spammer'
+          title: 'Mark %{name} as a verified not spammer'
```
notalex, line 149: **non spammer** might be more appropriate here than **not spammer**.

notalex (approved):
Minor changes. Everything is otherwise good here.

app/controllers/spam_controller.rb
BlackDuckDave: Fixed.

db/migrate/20220718210823_create_reviewed_not_spammer.rb
BlackDuckDave: Not exactly sure what you mean here, It is coming in oh. potential_spammers as a view. Is there another way I should be doing it?

config/locales/accounts.en.yml
BlackDuckDave: Fixed.

app/controllers/spam_controller.rb
BlackDuckDave: Also fixed this in the test.

PR #1650, opened 2022-08-22
OTWO-6775 Added tables for Openhub/Scan Integration project

Priya5:


notalex (approved):

sig-krasnick (commented):
I think for better performance we will need to add a gin index for the jsonb data field.  https://www.postgresql.org/docs/14/datatype-json.html#JSON-INDEXING

PR #1652, opened 2022-08-29
OTWO-6783 Cache DNS queries

notalex:


app/lib/fisbot/api_access.rb
```diff
@@ -22,7 +25,28 @@ def fis_public_url
 
   private
 
+  def fisbot_resolved_url
+    return URL if Rails.env.development?
```
sig-krasnick, line 29: Won't this fail the first time called because @@fis_ip_url is nil?

sig-krasnick (approved):

PR #1654, opened 2022-09-07
OTWO-6792 project api

Niharika1117:


app/controllers/api/v1/projects_controller.rb
```diff
@@ -0,0 +1,46 @@
...
+                                                                                 'branch' => params[:branch] } } }
+    @project.editor_account = current_user
+    @project.assign_editor_account_to_associations
+    @project.manages.new(account: current_user) if params[:managed_by_creator].to_bool
+    @project.save!
+    create_code_location_subscription if @project.enlistments.exists?
+  rescue StandardError => e
+    render json: "Bad params #{e}", status: :bad_request
+  end
+  # rubocop:enable Metrics/AbcSize, Metrics/MethodLength
```
notalex, line 23: Move lines 12-18 to a new method called *save_project*. That should take care of the rubocop warnings. 
We should try to avoid disabling rubocop wherever possible.

notalex (changes requested):

app/controllers/api/v1/projects_controller.rb
```diff
@@ -0,0 +1,46 @@
...
+
+  private
+
+  def project_params
+    params.require(:project).permit(
+      :name, :description, :vanity_url, :url, :download_url, :managed_by_creator, :coverity_project_id,
+      project_licenses_attributes: [:license_id],
+      enlistments_attributes: [code_location_attributes: %i[branch url scm_type]]
+    )
```
notalex, line 32: Project api will not be passing in params matching this hash. We won't have *enlistment_attributes*, *project_licenses_attributes* etc. Consider what the actual api call would look like.

app/controllers/api/v1/projects_controller.rb
```diff
@@ -0,0 +1,46 @@
...
+class Api::V1::ProjectsController < ApplicationController
+  include JWTHelper
+
+  skip_before_action :verify_authenticity_token
+  before_action :authenticate_jwt
+
+  # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
+  def create
+    populate_project_from_forge
+    @project.enlistments_attributes = { '0' => { 'code_location_attributes' => { 'scm_type' => 'git',
```
notalex, line 12: Although the ticket description vaguely mentions github, this api must support all SCMs. Hence rename `'git'` here to `params[:scm_type] || 'git'`.

notalex (changes requested):

app/controllers/api/v1/projects_controller.rb
notalex: The **ProjectBuilder** class can be used here instead. The initializer in ProjectBuilder must return `@project` and you can then use:
```
@project = ProjectBuilder.new(account, hsh)
```
Will need some testing to see if it fits. Looks like it would work.

notalex (changes requested):

app/controllers/api/v1/projects_controller.rb
```diff
@@ -0,0 +1,46 @@
...
+      enlistments_attributes: [code_location_attributes: %i[branch url scm_type]]
+    )
+  end
+
+  def populate_project_from_forge
+    match = Forge::Match.first(params[:url])
+    Timeout.timeout(Forge::Match::MAX_FORGE_COMM_TIME) { @project = match.project } if match
+  rescue Timeout::Error, OpenURI::HTTPError, URI::InvalidURIError
+    flash.now[:notice] = t('.forge_time_out', name: match.forge.name)
+  end
```
notalex, line 40: This is similar to the function in *projects_controller*. Move both functions to a common concern and refactor this method to accept an argument for *url*.
Avoid duplicating code wherever possible.

app/controllers/api/v1/projects_controller.rb
```diff
@@ -0,0 +1,46 @@
...
+    render json: "Bad params #{e}", status: :bad_request
+  end
+
+  private
+
+  def project_params
+    params.require(:project).permit(
+      :name, :description, :vanity_url, :url, :download_url, :managed_by_creator, :coverity_project_id,
+      project_licenses_attributes: [:license_id],
+      enlistments_attributes: [code_location_attributes: %i[branch url scm_type]]
```
notalex, line 23: * No branch, scm_type.
* Do not use nested attributes here. Expected a linear list of params and convert them in controller into nested attributes.
* Rename the first url to homepage_url.

app/controllers/api/v1/projects_controller.rb
```diff
@@ -0,0 +1,46 @@
...
+  include JWTHelper
+  include ProjectsHelper
+
+  skip_before_action :verify_authenticity_token
+  before_action :authenticate_jwt
+
+  def create
+    save_project
+    create_code_location_subscription if @project.enlistments.exists?
+  rescue StandardError => e
```
notalex, line 13: Use standard if else pattern here.

app/controllers/api/v1/projects_controller.rb
```diff
@@ -0,0 +1,46 @@
...
+      :name, :description, :vanity_url, :url, :download_url, :managed_by_creator, :coverity_project_id,
+      project_licenses_attributes: [:license_id],
+      enlistments_attributes: [code_location_attributes: %i[branch url scm_type]]
+    )
+  end
+
+  # rubocop:disable Metrics/AbcSize
+  def save_project
+    populate_project_from_forge(params[:url], true)
+    enlistments_attributes = [{ 'code_location_attributes' => { 'scm_type' => params[:scm_type] || 'git',
```
notalex, line 30: Use `project_params` instead of `params`.

app/helpers/projects_helper.rb
```diff
@@ -94,6 +95,17 @@ def more_badges_link
...
+  def populate_project_from_forge(url, api)
+    match = Forge::Match.first(url)
+    Timeout.timeout(Forge::Match::MAX_FORGE_COMM_TIME) { @project = match.project } if match
+  rescue Timeout::Error, OpenURI::HTTPError, URI::InvalidURIError
+    if api == true
+      render json: t('.forge_time_out', name: match.forge.name)
+    else
+      flash.now[:notice] = t('.forge_time_out', name: match.forge.name)
+    end
+  end
```
notalex, line 107: Setting instance variables for an object in one class and then consuming it in another is a bad code pattern. 
See [this](https://softwareengineering.stackexchange.com/a/214853) for more detailed explanation.
To keep it short, try passing local variables when possible. Instance variables seem clean but lead to unmanageable mess in the future. As a rule, using them in the same class is usually fine. Passing variables around seems ugly but it is better practice.

notalex: In the above code, pass project instead of setting `@project`. Also for the conditional, check if the following works:
```
   api ? render json: t('.forge_time_out', name: match.forge.name)
    : flash.now[:notice] = t('.forge_time_out', name: match.forge.name)
```
Not sure if rubocop will allow it.

notalex (changes requested):
We need test scenarios here for the following points specified on Jira:

> We have to ensure that a project with the same name doesn't exist. We might be able to return back the existing project assuming that the code location url and branch are the same.

> We will need to formulate a unique vanity_url. Prefixing the vanity_url with scan_* might help to eliminate duplicates

> Description will need to be filled in. We should be able to get that from the Github description.

app/helpers/projects_helper.rb
Niharika1117: I can change this one but it is having a dependency on UI and API. It might take time to work on this. After the change, I have to test everything in UI and API.

Can we skip this one for this PR?

I will take care on next PR.

app/controllers/api/v1/projects_controller.rb
```diff
@@ -0,0 +1,49 @@
...
+  private
+
+  def project_params
+    params.permit(:name, :vanity_url, :homepage_url, :repo_url, :coverity_project_id, :license_name)
+  end
+
+  def save_project
+    project = populate_project_from_forge(project_params[:repo_url], true)
+    create_params(project) if project
+    @project = ProjectBuilder.new(current_user, @project_params || {}).fetch_project
```
notalex, line 28: Dont set `@project` here. Set it in `create`.

app/core/project/project_builder.rb
```diff
@@ -12,6 +12,10 @@ def create
     @project
   end
 
+  def fetch_project
+    @project
+  end
```
notalex, line 17: Use a `att_reader :project` instead. Same for other PRs.

notalex (approved):

Niharika1117:
retest this please

PR #1657, opened 2022-09-14
OTWO-6805 Script for create scan  project

Niharika1117:


script/create_scan_project_in_ui.rb
```diff
@@ -0,0 +1,65 @@
...
+    puts e.message
+  end
+
+  def fix_url(url)
+    new_url = url.strip.chomp('/')
+    return new_url unless url.match('github.com')
+
+    new_url.sub(/\.git$/, '')
+           .sub(%r{^git://}, 'https://')
+           .sub(%r{^git@github.com:}, 'https://github.com/')
```
notalex, line 45: This operation is already done in Fisbot's **code_locations_controller**. Any specific reason to cleanup url here?

script/create_scan_project_in_ui.rb
```diff
@@ -0,0 +1,65 @@
...
+    ActiveRecord::Base.connection.execute("
+
+      INSERT INTO code_location_scan (code_location_id, scan_project_id) VALUES
+        ( #{code_location_id}, #{row['id']} );
+
+    ")
+  end
+
+  def insert_failed_url_data(row)
+    CSV.open('vendor/Scan_failed_url.csv', 'a+', write_headers: true, headers: @headers) do |writer|
```
notalex, line 59: Avoid using **Uppercase** in filenames.

script/create_scan_project_in_ui.rb
Niharika1117: CSV file is having URL which needs to be fixed before making calls. So I am using the same functionality here also.

Niharika1117:
retest this please

Niharika1117:
retest this please

app/controllers/api/v1/projects_controller.rb
```diff
@@ -11,6 +11,7 @@ def create
     save_project
     if @project.save
       create_code_location_subscription if @project.enlistments.exists?
+      render json: @project&.id, status: :created
```
notalex, line 14: There is no reason for `@project` to be nil here. Avoid unnecessary use of `&`.

app/helpers/projects_helper.rb
```diff
@@ -99,11 +99,7 @@ def populate_project_from_forge(url, api)
     match = Forge::Match.first(url)
     Timeout.timeout(Forge::Match::MAX_FORGE_COMM_TIME) { @project = match.project } if match
   rescue Timeout::Error, OpenURI::HTTPError, URI::InvalidURIError
-    if api == true
-      render json: t('.forge_time_out', name: match.forge.name)
-    else
-      flash.now[:notice] = t('.forge_time_out', name: match.forge.name)
-    end
+    flash.now[:notice] = t('.forge_time_out', name: match.forge.name) if api == false
```
notalex, line 102: Use `unless api`.

script/create_scan_project_in_ui.rb
notalex: What is the reason for CSV file urls to be cleaned before making the api call? Is it a requirement we specified somewhere?

app/helpers/jwt_helper.rb
```diff
@@ -5,11 +5,11 @@ def build_jwt(user, valid_for_hours = 48)
     exp = Time.now.to_i + (valid_for_hours * 60 * 60)
     payload = { expiration: exp,
                 user: user }
-    JWT.encode(payload, ENV['JWT_SECRET'], 'HS256')
+    JWT.encode(payload, ENV['JWT_SECRET_KEY'], 'HS256')
   end
 
   def decode_jwt(jwt)
-    decoded_token = JWT.decode(jwt, ENV['JWT_SECRET'], true)
+    decoded_token = JWT.decode(jwt, ENV['JWT_SECRET_KEY'], true)
```
notalex, line 12: JWT tokens are supposed to be generated before making api calls. Hence this key doesn't need to be in **kubernetes_config**.
* Generate the key before running this script. Export the `JWT_SECRET_KEY` in bash which will be read in the `ENV` here.
* Ask David Kronenberg on how to generate the key. It must be documented somewhere. If it is not, mention it on slack.
* Revert the kubernetes_openhub commit which added the jwt secret key.

app/controllers/api/v1/projects_controller.rb
```diff
@@ -22,11 +23,15 @@ def project_params
...
   end
 
+  # rubocop:disable Metrics/AbcSize
   def save_project
     populate_project_from_forge(project_params[:repo_url], true)
+    @project.name = params[:name] if @project && params[:name]
+    @project.coverity_project_id = params[:coverity_project_id] if @project && params[:coverity_project_id]
     create_params(@project) if @project
     @project = ProjectBuilder.new(current_user, @project_params || {}).fetch_project
   end
```
notalex, line 33: You can change the logic to the following which is the same behaviour. It will get rid of the metrics/abcsize warning:
```
    populate_project_from_forge(project_params[:repo_url], true)
    if @project
      @project.name = params[:name]
      @project.coverity_project_id = params[:coverity_project_id]
      create_params(@project)
    end
    ProjectBuilder.new(current_user, @project_params || {}).fetch_project
```

Also when u pass `project` as mentioned in the #1654 PR, stop using `@project` here and pass `project` back to calling method instead.

script/create_scan_project_in_ui.rb
```diff
@@ -0,0 +1,71 @@
...
+      next if project.nil?
+
+      puts "Project Value #{project} row value #{row['name']}"
+      if project.instance_of?(Hash) && (project['name'].to_s.include?("can't be blank") || project['enlistments.base'])
+        insert_failed_url_data(row)
+      end
+      create_code_location_scan(project, row) if project.is_a? Numeric
+    end
+  end
+  # rubocop:enable Metrics/AbcSize, Metrics/MethodLength
```
notalex, line 33: Avoid disabling `Metrics/MethodLength` since it is easy to fix.

app/models/forge/match.rb
```diff
@@ -19,12 +19,14 @@ def forge_id
...
     return {} unless json_api_url
 
-    # rubocop:disable Naming/MemoizedInstanceVariableName
-    @json ||= JSON.parse(URI.parse(json_api_url).open('User-Agent' => 'Ohloh.net client').read)
-    # rubocop:enable Naming/MemoizedInstanceVariableName
+    request = Net::HTTP::Get.new(json_api_url)
+    if ENV['GITHUB_AUTH_TOKEN'] && json_api_url.host == 'api.github.com'
+      request['Authorization'] = "token #{ENV['GITHUB_AUTH_TOKEN']}"
+    end
+    get_result(json_api_url, request)
```
notalex, line 29: Rename `json_api_url` to `json_api_uri`. Use `url` naming for strings and `uri` for **URI** objects.

app/models/forge/match.rb
```diff
@@ -43,6 +45,15 @@ def to_s
     owner_at_forge ? "#{forge.name}:#{owner_at_forge}/#{name_at_forge}" : "#{forge.name}:#{name_at_forge}"
   end
 
+  def get_result(uri, request)
+    http = Net::HTTP.new(uri.host, uri.port)
+    http.use_ssl = true
+    request['Content-Type'] = 'application/json'
+    request['User-Agent'] = 'Ohloh.net client'
```
notalex, line 52: Rename **Ohloh** to **openhub**.

fixtures/vcr_cassettes/ForgeMatchBitbucket.yml
```diff
@@ -71,4 +71,77 @@ http_interactions:
         true, "no_public_forks": false, "creator": null, "resource_uri": "/1.0/repositories/durin42/hgsubversion"}'
     http_version: 
   recorded_at: Fri, 20 Mar 2015 14:02:30 GMT
+- request:
```
notalex, line 74: Since you have new casettes, you may delete the older one above this. Run the specs and see if it works after removal.

script/create_scan_project_in_ui.rb
Niharika1117: If we are not cleaning the URL before making calls, it is not fetching the value from the repo and hence in the forge method, we are getting an empty result. We are not able to create a project.

app/controllers/api/v1/projects_controller.rb
```diff
@@ -24,10 +25,18 @@ def project_params
...
-    create_params(@project) if @project
+    assign_params(@project)
     @project = ProjectBuilder.new(current_user, @project_params || {}).fetch_project
   end
 
+  def assign_params(project)
+    return unless project
+
+    project.name = params[:name] if params[:name]
+    project.coverity_project_id = params[:coverity_project_id] if params[:coverity_project_id]
```
notalex, line 36: We don't need the `if` conditions here. `project.name` will be nil `if params[:name]` is nil. Same goes for coverity_project_id. There is no value in checking for the presence of these 2 values.

app/helpers/jwt_helper.rb
notalex: Discard this. We do need *JWT_SECRET_KEY* as an environment value. This is the secret key used to sign new JWT tokens.

app/controllers/api/v1/projects_controller.rb
notalex: Move this code back into `save_project`.

app/helpers/projects_helper.rb
```diff
@@ -94,6 +95,13 @@ def more_badges_link
     end
   end
 
+  def populate_project_from_forge(url, api)
+    match = Forge::Match.first(url)
+    Timeout.timeout(Forge::Match::MAX_FORGE_COMM_TIME) { @project = match.project } if match
```
notalex, line 100: Same as the other PR, avoid `@project` here.

notalex (approved):

PR #1659, opened 2022-09-20
OTWO-6810 Sbom link has been added to UI

Niharika1117:


app/controllers/project_sboms_controller.rb
```diff
@@ -0,0 +1,22 @@
...
+class ProjectSbomsController < ApplicationController
+  helper ProjectsHelper
+
+  before_action :find_project
+
+  def index
+    @project_sbom = @project.project_sboms&.first if @project
+    respond_to do |format|
+      format.js
+    end
```
notalex, line 12: Try removing lines 10-12 and check if it works.

app/controllers/project_sboms_controller.rb
```diff
@@ -0,0 +1,22 @@
...
+  end
+
+  private
+
+  def find_project
+    return nil unless params[:project_id]
```
notalex, line 18: Change to `return unless ...`.
nil is implicit.

app/decorators/project_decorator.rb
```diff
@@ -44,7 +44,8 @@ def sidebar
       [
         [:scm_data,         I18n.t(:scm_data)],
         [:commits,          I18n.t(:commits_menu),       h.summary_project_commits_path(project)],
-        [:contributors,     I18n.t(:contributors),       h.summary_project_contributors_path(project)]
+        [:contributors,     I18n.t(:contributors),       h.summary_project_contributors_path(project)],
+        [:sbom, project.project_sboms.count.positive? ? I18n.t(:sbom) : nil, h.project_project_sboms_path(project)]
```
notalex, line 48: Use the following instead:
```
 [:sbom, project.project_sboms.count.positive? && I18n.t(:sbom), h.project_project_sboms_path(project)]
```

app/models/concerns/project_associations.rb
```diff
@@ -41,6 +41,7 @@ module ProjectAssociations
     has_many :commit_flags, -> { order(time: :desc) }, through: :analysis_sloc_sets
     has_one :project_vulnerability_report
     has_many :commit_contributors
+    has_many :project_sboms # this might get changed
```
notalex, line 44: Remove the comment.

app/views/layouts/partials/_navigator.html.haml
```diff
@@ -17,4 +17,11 @@
         - links.each do |link|
           - link_class, link_name, link_url = link
           %li.footer-nav{ class: ('nav-active' if link_class == page_context[:select_footer_nav]) }
-            %a{ href: link_url }= link_name
+            - if link_name == 'SBOM'
+              = link_to link_name, link_url, remote: true,
+                    data: { toggle: 'modal', target: '#sbom_details.modal', keyboard: true }
+            - else
+              %a{ href: link_url }= link_name
+.modal.fade#sbom_details{ role: 'dialog', tabindex: -1 }
```
notalex, line 25: Use pattern like `sbom-details` for html *ID* names.

app/decorators/project_decorator.rb
notalex, line 48: I believe `project.project_sboms.exists?` will work here.

app/models/concerns/project_associations.rb
notalex: Change this to
```
has_many :project_sboms, as: :sboms
```

Use `project.sboms` instead of `project.project_sboms` elsewhere.

config/routes.rb
```diff
@@ -310,6 +310,7 @@
         post :redo
       end
     end
+    resources :project_sboms
```
notalex, line 313: Add `, only: :index`.

app/controllers/project_sboms_controller.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+class ProjectSbomsController < ApplicationController
+  helper ProjectsHelper
+
+  before_action :find_project
```
notalex, line 6: There is an function in `ApplicationController` named `set_project_or_fail`. Use it here instead.
Don't check for `if @project` in the `index` action, since `set_project_or_fail` will raise exception when `@project` is not set.

test/controllers/project_sboms_controller_test.rb
```diff
@@ -0,0 +1,30 @@
...
+      get :index, params: { project_id: 'dummy data' }
+      _(assigns(:project_sbom)).must_be_nil
+    end
+
+    it 'should return project nil for invalid project id' do
+      @controller = ProjectSbomsController.new
+      get :index, params: { project_id: 'dummmyyyyyyyyyy' }
+      @controller.instance_eval { find_project }
+      _(assigns(:project)).must_be_nil
+    end
```
notalex, line 28: Have one test showing a success scenario. It should assert that the view has the sbom rendered.
`response.body.must_match '{...}'`.

app/controllers/project_sboms_controller.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class ProjectSbomsController < ApplicationController
+  helper ProjectsHelper
+
+  before_action :set_project_or_fail
+
+  def index
+    @project_sbom = @project.sboms&.first
```
notalex, line 9: `sboms` will be an enumerable, which is never nil. `[].first` is always nil. So we don't need `&.` here.

app/views/project_sboms/_index.html.haml
```diff
@@ -0,0 +1,12 @@
+:ruby
+  sbom_data = project_sbom.sbom_data ?  project_sbom.sbom_data : {}
+
+.modal-header
+  %button.close{ data: { dismiss: 'modal' }, aria: { label: 'Close' } }
+    &times;
+  %h3.modal-title= "SBOM Details"
```
notalex, line 7: This text needs to be **i18n**.

notalex (changes requested):

PR #1672, opened 2022-10-31
OTWO-6848 SAST project 

Niharika1117:


app/views/scan_analytics/index.js.erb
```diff
@@ -0,0 +1,7 @@
+$("#scan_data").on("ajax:before", function() {
+  $('.overlay-loader').show();
+});
+$("#scan_data").html("<%= escape_javascript(render partial: 'index', locals: {scan_data: @scan_data, url: @analytics}) %>");
+$('.chzn-select').chosen();
+scanDataFetch();
+
```
notalex, line 7: Please remove the extra line here.

app/views/scan_analytics/index.html.haml
```diff
@@ -0,0 +1 @@
+= render partial: 'scan_analytics/index'
```
notalex, line 1: Configure your editor to automatically add newlines. If you are not sure, just open this file in vim and then type in `:x` to save and close the file. Vim will automatically add the newline for you.

test/controllers/scan_analytics_controller_test.rb
```diff
@@ -0,0 +1,53 @@
...
+                                                                              'components_loc' => 'N/A',
+                                                                              'defect_density' => { 'score' => 0.04,
+                                                                                                    'over_time' =>
+                                                                                                    [{ '2022-1-1' =>
+                                                                                                        0 }] },
+                                                                              'dismissed_count' => 0,
+                                                                              'prev_build_date' => '2022-10-13',
+                                                                              'eliminated_count' => 0,
+                                                                              'outstanding_count' => 35 } })
+  end
```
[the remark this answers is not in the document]
notalex: Like:
```
data = { data: { 'cwe' ..
@scan_analytics = create(:scan_analytic, .... , data: data)
```

app/views/scan_analytics/_index.html.haml
```diff
@@ -0,0 +1,67 @@
...
+  total_defects = scan_data.analysis_metrics.total_count
+  outstanding_count = scan_data.analysis_metrics.outstanding_count
+  fixed_count = scan_data.analysis_metrics.fixed_count
+#scan_data
+  .row.row-eq-height.project_row
+    %h4#scan_analytic.text-left Static Analysis
+  - if @analytics.size > 1
+    .row.row-eq-height.project_row
+      .pull-left#scan_analytic#scan_dropdown
+        = select_tag :code_set_id, options_for_select(@analytics.collect{ |u| [u.code_set.code_location.url, u.code_set_id] }, params[:code_set_id]), class: 'chzn-select value-select', data: { remote: true, url: project_scan_analytics_path(@project), method: 'get' }
```
notalex, line 18: This line is too long. Please split it over multiple lines.

app/controllers/scan_analytics_controlller.rb
```diff
@@ -0,0 +1,38 @@
...
+class ScanAnalyticsController < ApplicationController
+  helper ProjectsHelper
+
+  before_action :set_project_or_fail
+  before_action :analytics_data, only: :index
+
+  def index
+    return if @analytics.blank?
+
+    params[:code_set_id] = params[:code_set_id].nil? ? @analytics.first.code_set_id : params[:code_set_id]
```
notalex, line 12: This can be rewritten to:
```
  params[:code_set_id] = params[:code_set_id] ? params[:code_set_id] :  @analytics.first.code_set_id
```
We do not need to check for `nil`.

app/controllers/scan_analytics_controlller.rb
```diff
@@ -0,0 +1,38 @@
...
+    @scan_data = @analytics.where(code_set_id: params[:code_set_id]).first
+  end
+
+  def charts
+    return unless @project.coverity_project_id
+
+    scan_project_id = params[:scan_id].presence || @project.coverity_project_id
```
notalex, line 19: I believe you do not need `.presence` here.

app/controllers/scan_analytics_controlller.rb
```diff
@@ -0,0 +1,38 @@
...
+  def analytics_data
+    @analytics = @project.best_analysis&.scan_analytics&.order(id: :asc) if @project.best_analysis
+  end
+
+  def api_response(scan_project_id)
+    uri = URI.parse("#{ENV['COVERITY_SCAN_URL']}/api/projects/#{scan_project_id}/charts?format=json")
+    http = Net::HTTP.new(uri.host, uri.port)
+    http.use_ssl = true
+    response = http.get(uri.request_uri)
+    raise NonHTTPSuccessResponse unless response.code == '200'
```
notalex, line 34: Will this exception break the UI for our users?

app/views/scan_analytics/_index.html.haml
Niharika1117: It's written in multiple lines in the editor. I don't know why it's displaying here as a single line.
Screenshot 2022-11-02 at 10 09 33 PM

app/controllers/scan_analytics_controlller.rb
Niharika1117: No, It will not break UI. Chart data will not load.

app/views/scan_analytics/_index.html.haml
notalex: Seems like your editor auto `wraps` it for you. There is a setting for it. Consider disabling the setting.

app/controllers/scan_analytics_controlller.rb
Niharika1117: params[:scan_id] it might be " " in params. SO handling that scenario is required.

app/assets/javascripts/scan_analytics.js
```diff
@@ -0,0 +1,210 @@
...
+          }
+        },
+        yAxis: {
+          title: {
+            text: 'Outstanding defects'
+          }
+        },
+      };
+      chart4options = jQuery.extend(true, {}, options, chart4options);
+      var chart4 = new Highcharts.Chart(chart4options);
```
notalex, line 207: Where is `chart4` used. If this variable is not used anywhere, remove it from here. Just call `new Highcharts`.

app/controllers/scan_analytics_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+    return if @analytics.blank?
+
+    params[:code_set_id] = params[:code_set_id] || @analytics.first.code_set_id
+    @scan_data = @analytics.where(code_set_id: params[:code_set_id]).first
+  end
+
+  def charts
+    return unless @project.coverity_project_id
+
+    scan_project_id = params[:scan_id].presence || @project.coverity_project_id
```
notalex, line 19: Like mentioned before, one doesn't need to check for `presence` of params. They are nil by default.

app/controllers/scan_analytics_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+    api_response(scan_project_id)
+  end
+
+  private
+
+  def analytics_data
+    @analytics = @project.best_analysis&.scan_analytics&.order(id: :asc) if @project.best_analysis
```
notalex, line 26: `scan_analytics` is an array. When there are no results, it will be an empty array, which will not raise an exception for `order`. Hence we do not need `&.` after `scan_analytics`. One never needs to use `&.` with an array.

notalex: Why are we doing `if @project.best_analysis` while also doing `@project.best_analysis` with an `&.`? You need either one of those, not both.

app/views/scan_analytics/_no_scan_analytics.html.haml
```diff
@@ -0,0 +1,7 @@
+.row.row-eq-height.project_row
+  %h4#scan_analytic.text-left Static Analysis
```
notalex, line 2: Move text to `config/locales`.

app/views/scan_analytics/_index.html.haml
```diff
@@ -0,0 +1,69 @@
...
+  version = scan_data.analysis_metrics.version.present? ? scan_data.analysis_metrics.version : nil
+  last_analysed = scan_data.analysis_metrics.build_date
+  loc = scan_data.analysis_metrics.loc
+  defect_density = scan_data.analysis_metrics.defect_density&.score
+  total_defects = scan_data.analysis_metrics.total_count
+  outstanding_count = scan_data.analysis_metrics.outstanding_count
+  fixed_count = scan_data.analysis_metrics.fixed_count
+#scan_data
+  .row.row-eq-height.project_row
+    %h4#scan_analytic.text-left Static Analysis
```
notalex, line 14: Move to `locales`.

app/views/scan_analytics/_index.html.haml
```diff
@@ -0,0 +1,69 @@
+:ruby
+  scan_data = @scan_data.present? ? @scan_data : scan_data
```
notalex, line 2: `scan_data` is never passed to this partial. So this line doesn't make sense.

app/views/scan_analytics/_index.html.haml
```diff
@@ -0,0 +1,69 @@
...
+  scan_data = JSON.parse(scan_data.data.to_json, object_class: OpenStruct)
+  @cwes = scan_data.cwe
```
notalex, line 4: Where does `scan_data.cwe` come from? `scan_data` is an `ScanAnalytics` object. The model or the database table doesn't have this column.

app/assets/javascripts/scan_analytics.js
```diff
@@ -0,0 +1,210 @@
...
+  $.ajax({
+    type: 'GET',
+    url: '/p/' + $('#chart_data').data('project-id') + '/scan_analytics/charts?scan_id=' + $('#chart_data').data('scan-project-id'),
+    success: function(data) {
+      var options = {
+        chart: {
+          zoomType: 'xy'
+        },
+      };
+      Highcharts.setOptions({
```
notalex, line 14: Move out logic into independent functions. Right now we are doing a bunch of different operations in the same block. It would be better to segregate it into well named functions.

app/views/scan_analytics/_index.html.haml
```diff
@@ -0,0 +1,69 @@
...
+  version = scan_data.analysis_metrics.version.present? ? scan_data.analysis_metrics.version : nil
```
notalex, line 5: Change this to
```
version = scan_data.analysis_metrics.version.presence
```

app/views/scan_analytics/_index.html.haml
```diff
@@ -0,0 +1,69 @@
...
+  last_analysed = scan_data.analysis_metrics.build_date
+  loc = scan_data.analysis_metrics.loc
+  defect_density = scan_data.analysis_metrics.defect_density&.score
+  total_defects = scan_data.analysis_metrics.total_count
+  outstanding_count = scan_data.analysis_metrics.outstanding_count
+  fixed_count = scan_data.analysis_metrics.fixed_count
```
notalex, line 11: Instead of assigning 7 variables here, assign a single variable `metrics = scan_data.analysis_metrics` and use `metrics.loc` etc. further down.

app/views/scan_analytics/_index.html.haml
notalex, line 4: Instead of instance variable, use a local variable for `cwes` and pass it to the partial.

app/controllers/scan_analytics_controller.rb
notalex: when `params[:scan_id` is present, is it equal to `@project.coverity_project_id`? If so, can we use only  `@project.coverity_project_id` here?

app/controllers/scan_analytics_controller.rb
Niharika1117: I got an rubocop error after removing & after scan_analytics.
Below is the error I got it
https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Lint/SafeNavigationChain

app/assets/javascripts/scan_analytics.js
```diff
@@ -0,0 +1,230 @@
...
+      }
+    },
+    yAxis: {
+      title: {
+        text: 'Outstanding defects'
+      }
+    },
+  };
+  chart4options = jQuery.extend(true, {}, options, chart4options);
+  new Highcharts.Chart(chart4options);
```
notalex, line 229: rename `chart4options` to `chart4Options`. Same for other similar variable names.

PR #1674, opened 2022-11-02
OTWO-6846 Updated script for creating scan projects in UI

Priya5:


app/controllers/api/v1/projects_controller.rb
```diff
@@ -24,31 +25,43 @@ def project_params
...
   end
 
   def create_params(project)
-    enlistments_attributes = [{ 'code_location_attributes' => { 'scm_type' => project.code_location_object.scm_type \
-                                                                              || 'git',
-                                                                'url' => project.code_location_object.url,
-                                                                'branch' => project.code_location_object.branch \
-                                                                          || 'main' } }]
+    project.name = params[:name] if params[:name]
+    project.vanity_url = params[:vanity_url] if params[:vanity_url]
```
notalex, line 35: Maybe we do not need these if conditions here. `params` will be nil by default.

notalex (approved):

PR #1678, opened 2022-11-22
OTWO 6874 mass assignment fix

Niharika1117:


app/controllers/helpfuls_controller.rb
```diff
@@ -4,14 +4,16 @@ class HelpfulsController < ApplicationController
   before_action :session_required, :redirect_unverified_account
 
   def create
-    @helpful = Helpful.where(model_params).first_or_initialize
+    @helpful = Helpful.where(account_id: params[:helpful][:account_id],
+                             review_id: params[:review_id]).first_or_initialize
+    @helpful.account = current_user
```
sig-krasnick, line 9: Since we are using the current_user here, we should be able to eliminate the account_id completely and find the correct record by using current_user.id instead.

app/controllers/widgets_controller.rb
```diff
@@ -19,7 +19,7 @@ def record_not_found
   def set_widget
     widget_name = action_name.split('_') - WIDGET_TYPES
     @widget = Object.const_get("#{controller_name.camelize[0..-2]}::#{widget_name.join('_').camelize}")
-                    .new(permitted_params)
+                    .new(permitted_params.to_h.merge(account_id: params[:account_id]))
```
notalex, line 22: This will not fix the original issue. See the reason for this change in the ticket.

PR #1681, opened 2022-11-23
OTWO-6875 Sql injection Fix

Niharika1117:


Priya5 (approved):

app/controllers/api/v1/enlistment_controller.rb
```diff
@@ -54,6 +54,7 @@ def find_project
   end
 
   def delete_all_subscriptions(code_location_id)
-    Enlistment.connection.execute("DELETE FROM fis.subscriptions WHERE code_location_id =#{code_location_id};")
+    Enlistment.connection.execute("DELETE FROM fis.subscriptions WHERE code_location_id =
+                                  #{ActiveRecord::Base.sanitize_sql(code_location_id)};")
```
notalex, line 58: This `code_location_id` comes directly from the DB. We shouldn't need to sanitize it here. You may undo this change.
Is there any way to mark this line as safe for coverity scan?

app/controllers/edits_controller.rb
```diff
@@ -84,7 +84,9 @@ def add_where_term(edits)
     target_where = '(edits.target_id = ? AND edits.target_type = ?)'
     extra_where = add_where_extra_clause
     if extra_where
-      edits.where(["#{target_where}#{extra_where}", @parent.id, @parent.class.name.tableize, @parent.id])
+      edits.where(ActiveRecord::Base.send(:sanitize_sql_array, ["#{target_where}#{extra_where}",
+                                                                @parent.id, @parent.class.name.tableize,
+                                                                @parent.id]))
```
notalex, line 89: This change is also not required.

notalex (changes requested):
Please review all the changes here and only sanitize values that come directly from end users.

PR #1684, opened 2022-12-06
OTWO-6892 Admin permission to sbom

Niharika1117:


app/decorators/project_decorator.rb
```diff
@@ -26,7 +26,7 @@ def sorted_link_list
   end
 
   # rubocop:disable Metrics/MethodLength, Metrics/AbcSize
-  def sidebar
+  def sidebar(current_user)
```
notalex, line 29: Rename `current_user` to `account` because the actual object is account. Variable naming should reflect intention. Here the intention is passing an `account` object, it doesn't matter if the account object is current_user or some other user.

test/decorators/project_decorator_test.rb
```diff
@@ -4,6 +4,14 @@
 
 class ProjectDecoratorTest < ActiveSupport::TestCase
   let(:linux) { create(:project) }
+  before do
+    admin = create(:admin)
+    stubs(:current_user).returns(admin)
+    @enlistments = create(:enlistment, project: linux, code_location_id: 1)
+    @project_sboms = create(:project_sbom, agent: 'syft', project_id: linux.id,
+                                           code_location_id: @enlistments.code_location_id,
+                                           sbom_data: '{\"SPDXID\":\"SPDXRef-DOCUMENT\"}')
```
notalex, line 13: Change this line to:
```
  sbom_data: { SPDXID: 'SPDXRef-DOCUMENT' }.to_json)
```
which is cleaner.

test/decorators/project_decorator_test.rb
```diff
@@ -4,6 +4,14 @@
 
 class ProjectDecoratorTest < ActiveSupport::TestCase
   let(:linux) { create(:project) }
+  before do
+    admin = create(:admin)
+    stubs(:current_user).returns(admin)
```
notalex, line 9: This is equivalent to writing `current_user = admin`. It doesn't stub the actual `current_user` in ProjectDecorator. Remove this line and use `admin` instead of `current_user` in the specs.

test/decorators/project_decorator_test.rb
```diff
@@ -30,16 +38,18 @@ class ProjectDecoratorTest < ActiveSupport::TestCase
...
 
   describe 'sidebar' do
     it 'should contain 4 sections' do
-      _(linux.decorate.sidebar.length).must_equal 4
+      _(linux.decorate.sidebar(current_user).length).must_equal 4
     end
 
     it 'should return projects menu list' do
-      _(linux.decorate.sidebar).must_equal sidebar
+      _(linux.decorate.sidebar(current_user)).must_equal sidebar
```
notalex, line 52: This test is incorrect. Think about what we are doing here.
On the left hand side, we are creating a sidebar object with the same code that we see in the ProjectDecorator. On the right hand side, we call similar code in ProjectDecorator.
The original test was also incorrect, so this is not something you need to deal with in this PR. This is just to be aware of bad patterns in tests.

test/helpers/page_context_helper_test.rb
```diff
@@ -64,6 +64,8 @@ class PageContextHelperTest < ActionView::TestCase
   end
 
   it 'should return project page context' do
+    @account = create(:admin)
+    stubs(:current_user).returns(@account)
```
notalex, line 68: This is the same as writing : `current_user = @account`. If you remove this line, this test should behave the same. Try running the test without these lines.

notalex (changes requested):

test/helpers/page_context_helper_test.rb
Niharika1117: stubs(:current_user).returns(@account) removing this line causing test case failure. So , I am not making this change.

notalex: What is the error? Always mention the actual error instead of stating that something causes an error.

notalex: If there is no error and this is simply fixing the test, then we should know how this fixes the test. 
Add a byebug in the controller and check if `current_user` in the controller is the same as `@account` from the test. This would mean that this stubs line actually is stubbing inside the controller.

PR #1695, opened 2023-01-16
OTWO-6916 Add DB backed code_location model

Niharika1117:


test/factories/fis_code_location.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :fis_code_location do
+    is_important { true }
+    do_not_fetch { true }
```
notalex, line 6: Avoid these default boolean values in this factory because they seem to cater to the specific needs of `project_test` spec. Set these flags in the spec itself, which makes the spec clearer.

test/models/project_test.rb
```diff
@@ -543,4 +543,24 @@ class ProjectTest < ActiveSupport::TestCase
       _(project.searchable_vector[:d]).must_equal project.description
     end
   end
+
+  describe 'important projects' do
+    it 'should return important code locations project' do
+      enlist = create(:enlistment, project: project)
+      fis_code_location = FisCodeLocation.first
```
notalex, line 550: Change the above to :
```
enlistment = create(:enlistment, project: project)
fis_code_location = enlistment.fis_code_locations.last
```

Rename the variable elsewhere. Same change for the other test.

test/models/project_test.rb
```diff
@@ -543,4 +543,24 @@ class ProjectTest < ActiveSupport::TestCase
...
+      fis_code_location.update(is_important: true)
+      enlist.code_location_id = fis_code_location.id
```
notalex, line 552: This should already be pointing to `fis_code_location`. Try removing this line.

notalex (approved):

PR #1698, opened 2023-01-19
OTWO-6574 fix vulnerability issue for empty data

Niharika1117:


app/helpers/vulnerabilities_helper.rb
```diff
@@ -55,7 +55,7 @@ def options_for_severities_filter
   end
 
   def sort_releases_by_version_number(releases)
-    return nil if releases.nil?
+    return nil if releases.blank?
```
notalex, line 58: Change this to
```
return if releases.blank?
```

PR #1701, opened 2023-01-24
OTWO-6926 active enlistment filter issue

Niharika1117:


sig-krasnick (approved):
Tested on staging.  Working as expected.

PR #1716, opened 2023-03-14
OTWO-6954 api for create scan project

Niharika1117:


app/controllers/api/v1/projects_controller.rb
```diff
@@ -63,4 +71,14 @@ def code_location_branch(url)
     out, _err, _status = Open3.capture3("git ls-remote --symref #{url} HEAD | head -1 | awk '{print $2}'")
     out.strip.sub(/refs\/heads\//, '')
   end
+
+  def get_scan_api_data(url, path)
+    project = populate_project_from_forge(url, true)
```
sig-krasnick, line 76: I'm a bit confused by this line.  Why do we have to go to the forge?  The project already exists on Openhub.  Don't we need to pass in the codelocation.repository.url to the scan api?

Luna88k (approved):

PR #1728, opened 2023-06-14
OTWO-6997 duplicate resolve was failing

Niharika1117:


app/models/enlistment.rb
```diff
@@ -12,7 +12,7 @@ class Enlistment < ApplicationRecord
 
   before_save :save_code_location, if: -> { @nested_code_location }
   after_update :update_subscription, if: :saved_change_to_deleted?
-  after_save :ensure_forge_and_job
+  after_save :ensure_forge_and_job, unless: ->(enlistment) { enlistment.deleted? }
```
notalex, line 15: Try changin to the following and check if it works:
```
after_save :ensure_forge_and_job, unless: :deleted?
```

notalex (approved):

PR #1730, opened 2023-06-16
OTWO-7024 Delete project without enlistments

Niharika1117:


Niharika1117:
Done the work as expected

PR #1738, opened 2023-08-02
OTWO-7056 Allow customers to report about outdated project

Priya5:


app/mailers/project_mailer.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+class ProjectMailer < ApplicationMailer
+  def report_outdated(account, project)
+    @project = project
+    mail to: 'info@openhub.net', from: account.email, subject: "#{@project.name} is outdated"
```
notalex, line 6: The `from` option might not work. Our sendgrid is configured to allow sending from any openhub.net email address. Change `from` to openhub.net and add the user email in the html.

notalex (approved):

PR #1756, opened 2023-11-20
OTWO-7139 Fix unique violation for positions

Niharika1117:


app/controllers/positions_controller.rb
```diff
@@ -89,4 +91,9 @@ def position_params
                   :affiliation_type, :description, :start_date, :stop_date, :ongoing, :invite,
                   language_exp: [], project_experiences_attributes: %i[project_name _destroy id])
   end
+
+  def find_project
+    @project = Project.not_deleted.find_by('lower(name) = ? or name = ?', position_params['project_oss'].to_s.downcase,
+                                           position_params['project_oss'])
+  end
```
alex-sig, line 98: Do we need to check both `lower(name)` and `name`?
I believe only `lower(name)` would suffice.

alex-sig (approved):

PR #1758, opened 2023-12-14
OTWO-7156 Handle more vulnerability version patterns

Niharika1117:


app/assets/javascripts/vulnerabilities.js.coffee
```diff
@@ -104,7 +104,7 @@ filterReleasesByMajorVersion = (releases, majorVersion) ->
       ///^#{majorVersion}///.test(release.version)
   else
     releases.filter (release) ->
-      ///^#{majorVersion}+(?:\.\d+)+$///.test(release.version)
+      ///^(v)+#{majorVersion}+(?:\.\d+)+$///.test(release.version)
```
alex-sig, line 107: Change `(v)+` to `v?`. 
`?` means `v` is optional and `()` is used for optional matches or when we are going to use the value after matching.

app/helpers/vulnerabilities_helper.rb
```diff
@@ -11,7 +11,7 @@ def major_releases(releases, project)
       releases.pluck(:version)
     else
       releases.map do |r|
-        r[:version].to_i
+        r[:version].gsub(/[^\d,.]/, '').to_i
```
alex-sig, line 14: This logic will break when the pattern is `v.2.2.2`.
Change regex to `/^v?\.?/`. 
When changing regexp, always play around with it in irb with different inputs first.
Also add include this pattern in vulnerabilities_helper_test.

alex-sig (approved):

PR #1783, opened 2024-04-03
OTWO-7190 Add primary key for commits

Priya5:


alex-sig:
Not required anymore.

PR #1815, opened 2025-03-30
views: fix spelling in `_footer.html.haml`

jbampton:


alex-sig:
@jbampton, thanks for the fix. This will be merged & deployed next week.

PR #1819, opened 2025-05-02
OTWO-7324 Report missing form params

Niharika1117:


app/controllers/accounts_controller.rb
```diff
@@ -118,6 +120,11 @@ def account_params
     )
   end
 
+  def handle_parameter_missing(exception)
+    flash[:error] = "Required parameter is missing: #{exception.param}"
+    redirect_to new_account_path
```
alex-sig, line 125: Let us not throw this flash error. It will expose the issue to the exploiter, which will allow them to understand why it failed. We do not want that.
Instead send the error to errbit and just continue the redirect to the new accounts page without any flash error.

Niharika1117:
retest this

alex-sig (approved):

PR #1877, opened 2026-03-20
OTWO-7585 Fix invalid byte sequence in UTF-8

Niharika1117:


app/helpers/links_helper.rb
```diff
@@ -4,7 +4,9 @@
 
 module LinksHelper
   def safe_slice_host(url, length = 33)
-    safe_url = sanitize(CGI.unescape(url))
+    # Unescape and force valid UTF-8 before sanitizing
+    unescaped = CGI.unescape(url).encode('UTF-8', invalid: :replace, undef: :replace, replace: '')
+    safe_url = ActionController::Base.helpers.sanitize(unescaped).to_s
```
Priya5: Instead of these changes above, replace line 7 with the following, as we already have this fix_encoding_if_invalid method for String:
`safe_url = sanitize(CGI.unescape(url).fix_encoding_if_invalid)`

There is no need to use ActionController::Base.helpers.sanitize in this context, since helper modules already have direct access to the sanitize method.

Priya5 (approved):

PR #1890, opened 2026-04-30
OTWO-7608 Add docker setup for ohloh-ui

ashutok036221:
Created Dockerfile and compose along with make file for developers to run ohloh-ui as docker services on local.
DOCKER.md can be referred to how to use docker command to run/stop application.
Readme.md is also updated to have reference to DOCKER.md


.env
```diff
@@ -1,3 +1,7 @@
+# Docker build settings - disable BuildKit (causes hangs with docker-compose v1)
+DOCKER_BUILDKIT=0
```
Priya5, line 2: This is not required as it is exported in the Makefile.

PR #1895, opened 2026-05-11
OTWO-7639: added security policy to allow cross origin resources for whitelist domains

ashutok036221:
Since, OpenHUB widgets are not working for some customers, added security policy to allow cross origin resources for whitelisted domains to allow widgets to load.
For other customers using widgets we need to add the domain to allow them too.


Priya5:
Since this issue was caused by the user's CSP not allowing https://openhub.net and the Cloudflare-managed captcha, this change is no longer necessary.

PR #1905, opened 2026-06-05
OTWO-7667 People page Query optiization

bd-vaibhav:


Niharika1117 (approved):

app/controllers/people_controller.rb
```diff
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 class PeopleController < UnclaimedController
+  include ApplicationHelper
```
Priya5: Please remove `include ApplicationHelper` from the controller. Helpers belong in views and including the entire module introduces unrelated methods into the controller namespace.
Add the `only_device?` method directly into the controller as a private method.

Priya5 (approved):

PR #1915, opened 2026-06-25
Check for presence of table index

Priya5:


Priya5:
This is not required anymore. hence, closing this.

PR #1923, opened 2026-07-21
OTWO-7703 Password Policy change to minimum 12 characters

bd-vaibhav:


app/models/concerns/account_validations.rb
```diff
@@ -7,7 +7,8 @@ module AccountValidations
     validates :email, presence: true, length: { in: 3..100 }, uniqueness: { case_sensitive: false },
                       email_format: true, allow_blank: false
 
-    validates :password, length: { in: 5..40 }, unless: :skip_password_validation?
+    validates :password, length: { in: 12..64 }, allow_blank: true, unless: :skip_password_validation?
```
Priya5: why are we adding `allow_blank` ?

bd-vaibhav: because in a case no password is provided. Blank password validation runs and rejects and then shows error message. If we don't allow blank then 2 error messages are shown together like 'Please provide a password' & 'password should be atleast 12 characters long'. But we only need to show one error message at a time.

Priya5 (approved):

PR #1941, opened 2026-08-12
OTWO-7728 Server Reflects X-Forwarded-Host Header Value in API Responses

Niharika1117:


Niharika1117:
Not required I have checked this is already working

